PR #8, opened 2020-02-10
Add Settings Page

nicholas-cook:
- Adds settings page for viewing privacy policy and contacting developers
- Added new button classes to handle close and external linking functionality
- Hooked settings page up to main page


app/src/main/res/layout/activity_settings.xml
```diff
@@ -0,0 +1,122 @@
...
+        app:layout_constraintBottom_toTopOf="@id/privacy_policy_button"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintVertical_chainStyle="packed" />
+
+    <com.example.eyespeak.ActionButton
+        android:id="@+id/privacy_policy_button"
+        android:layout_width="0dp"
+        android:layout_height="144dp"
```
CarolineLaw: we could pull some of this into a style since we're duplicating some of it

app/src/main/java/com/example/eyespeak/ActionButton.kt
```diff
@@ -0,0 +1,45 @@
...
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyle: Int = 0
+) : EyeSpeakButton(context, attrs, defStyle) {
+
+    var action: (() -> Unit)? = null
+
+    init {
+        setOnClickListener {
+            Toast.makeText(context, text, Toast.LENGTH_LONG).show()
```
jequeen: We are duplicating logic in our custom button classes, we should factor that out eventually.

app/src/main/res/layout/activity_settings.xml
nicholas-cook: We can address this in our cleanup ticket.

app/src/main/java/com/example/eyespeak/ActionButton.kt
nicholas-cook: We can address this in our cleanup ticket.

PR #10, opened 2020-02-10
Refactor buttons and update project structure

nicholas-cook:
- Moves common base button functionality to the main `VocableButton` class
- Updates face processing job in the view model to skip frames during processing
- Moves classes into appropriate packages instead of the main base package
- Adds string resource values for button text
- Changes the app package from `com.example.eyespeak` to `com.willowtree.vocable`


app/src/main/java/com/willowtree/vocable/customviews/VocableButton.kt
```diff
@@ -0,0 +1,75 @@
...
+import android.content.Context
+import android.speech.tts.TextToSpeech
+import android.util.AttributeSet
+import androidx.appcompat.widget.AppCompatButton
+import com.willowtree.vocable.utils.VocableTextToSpeech
+import kotlinx.coroutines.*
+
+/**
+ * A custom AppCompatButton that will delay for two seconds when a pointer enters and then will call
+ * VocableTextToSpeech to say the button text allow and then perform an action based on the
```
jequeen: "allow" shouldn't be here.

PR #14, opened 2020-02-14
Fragment refactor

nicholas-cook:
- Refactors the app to separate functionality into fragments instead of having `MainActivity` responsible for all behavior
- This is to make it easier to load content more dynamically in the future and to prevent any one view class from becoming too monolithic. Hopefully we can divide behavior out more granularly using additional fragments and view models.
- Also adds basic keyboard screen (designs to change later) and switching between phrases and keyboard

### DIFF
--- a/app/src/main/res/layout/fragment_keyboard.xml
+++ b/app/src/main/res/layout/fragment_keyboard.xml
@@ -1,0 +1,247 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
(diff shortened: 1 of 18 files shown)


app/src/main/res/layout/fragment_keyboard.xml
```diff
@@ -0,0 +1,255 @@
...
+            <com.willowtree.vocable.customviews.ActionButton
+                android:layout_width="0dp"
+                android:layout_height="104dp"
+                android:layout_weight="1"
+                android:background="@drawable/radio_button_default_background"
+                android:gravity="center" />
+
+        </LinearLayout>
+
+        <LinearLayout
```
jequeen: We are going to change this eventually to have single key layout. We may use this layout for a future "two-stroke" keyboard.

jequeen: When we re-visit this layout we can also add proper styles and pull out dimensions and strings.

PR #15, opened 2020-02-17
Implement single stroke keyboard

nicholas-cook:
- Updates keyboard layout to use single keys instead of groups of keys
- Adds logic to set keys dynamically instead of hardcoding them via xml
- Adds logic to pass text from keys to EditText
- Adds speak, clear, and backspace functionality to keyboard


app/src/main/res/layout/fragment_keyboard_2_stroke_ff.xml
```diff
@@ -0,0 +1,255 @@
+<?xml version="1.0" encoding="utf-8"?>
```
CarolineLaw: why are we keeping this?

nicholas-cook: We're keeping this layout around in case we go back to a two-stroke layout

PR #17, opened 2020-02-19
Add pagination for categories and phrases

nicholas-cook:
- Adds pagination functionality for the categories and phrases on the presets screen
- Uses ViewPager2s and fragments to accomplish pagination
- NOTE - Pagination only works for swiping and not head tracking

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
@@ -5,20 +5,30 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.core.view.children
 import androidx.core.view.isVisible
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentManager
 import androidx.lifecycle.Observer
+import androidx.lifecycle.ViewModelProviders
+import androidx.viewpager2.adapter.FragmentStateAdapter
 import com.willowtree.vocable.BaseFragment
 import com.willowtree.vocable.R
 import com.willowtree.vocable.customviews.PointerListener
 import com.willowtree.vocable.utils.SpokenText
 import com.willowtree.vocable.utils.VocableTextToSpeech
 import kotlinx.android.synthetic.main.fragment_presets.*
+import kotlin.math.ceil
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
```diff
@@ -53,4 +112,39 @@ class PresetsFragment : BaseFragment() {
...
+    }
+
+    inner class PhrasesPagerAdapter(fm: FragmentManager, private val phrases: List<String>) :
+        FragmentStateAdapter(fm, viewLifecycleOwner.lifecycle) {
+
+        override fun getItemCount(): Int {
+            return ceil(phrases.size / MAX_PHRASES.toDouble()).toInt()
+        }
+
+        override fun createFragment(position: Int): Fragment {
```
CarolineLaw: we should have infinite scrolling eventually :]

app/src/main/java/com/willowtree/vocable/presets/PresetsViewModel.kt
```diff
@@ -0,0 +1,242 @@
...
+        CATEGORY_CONVERSATION,
+        CATEGORY_ENVIRONMENT,
+        CATEGORY_FEELINGS,
+        CATEGORY_QUESTIONS,
+        CATEGORY_TIME,
+        CATEGORY_PEOPLE,
+        CATEGORY_NUMBERS
+    )
+
+    private val generalPhrases = listOf(
```
jequeen: In the future we could put our categories and phrase lists into JSON and parse those maybe.

app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
nicholas-cook: That's coming in the next PR!

app/src/main/java/com/willowtree/vocable/presets/PresetsViewModel.kt
nicholas-cook: Yeah I think that will definitely clean up this view model 👍

PR #21, opened 2020-02-20
Add a way to store presets

nicholas-cook:
- Adds Koin dependency injection
- Adds a `SharedPreferences` implementation we can use to save users' sayings
- Updates the `PresetsViewModel` so we can get/set users' sayings

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/MainActivity.kt
+++ b/app/src/main/java/com/willowtree/vocable/MainActivity.kt
@@ -5,12 +5,14 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.annotation.LayoutRes
 import androidx.core.view.children
+import androidx.lifecycle.ViewModelProviders
 import com.willowtree.vocable.customviews.ActionButton
 import com.willowtree.vocable.customviews.PauseButton
 import com.willowtree.vocable.customviews.PointerListener
 import com.willowtree.vocable.customviews.PointerView
 import com.willowtree.vocable.keyboard.KeyboardFragment
 import com.willowtree.vocable.presets.PresetsFragment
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/willowtree/vocable/MainActivity.kt
```diff
@@ -32,6 +35,8 @@ class MainActivity : BaseActivity() {
         keyboard_button.action = {
             loadFragment(KeyboardFragment())
         }
+
+        presetsViewModel = ViewModelProviders.of(this).get(PresetsViewModel::class.java)
```
CarolineLaw: might run into merge conflicts here, but that's okay :]

app/src/main/java/com/willowtree/vocable/presets/PresetsViewModel.kt
```diff
@@ -231,6 +237,11 @@ class PresetsViewModel : ViewModel() {
     val currentPhrases: LiveData<List<String>> = liveCurrentPhrases
 
     init {
+        val mySayings = sharedPrefs.getMySayings()
+        if (mySayings.isNotEmpty()) {
```
CarolineLaw: Does this remove My Sayings from the category list if it's empty? I feel like we should have an empty state instead of just not including it. We can add that in later though :]

nicholas-cook: Just talked to Kayce and I think she's going to add an empty state to the designs!

app/src/main/java/com/willowtree/vocable/MainActivity.kt
nicholas-cook: This was just a placeholder for now. It can be changed if needed.

PR #23, opened 2020-02-25
Update category and keyboard screens to match designs

nicholas-cook:
- Updated layouts of category and keyboard screens to match new design layouts
- Each fragment is now responsible for its own top bar action buttons instead of the activity
- Added a way to add centered icons to buttons in the base `VocableButton` class


app/src/main/java/com/willowtree/vocable/customviews/ActionButton.kt
```diff
@@ -22,8 +22,10 @@ open class ActionButton @JvmOverloads constructor(
 
     override fun sayText(text: CharSequence?) {
         text?.let {
-            VocableTextToSpeech.getTextToSpeech()
-                ?.speak(it, TextToSpeech.QUEUE_FLUSH, null, id.toString())
+            if (it.isNotBlank()) {
```
MattCorbett488: Could potentially get rid of the `.let` and condense this to `if (text?.isNotBlank() == true)` which would cover both your false and null cases

app/src/main/res/layout/fragment_keyboard.xml
```diff
@@ -7,34 +7,50 @@
...
         android:id="@+id/speaker_icon"
         android:layout_width="wrap_content"
-        android:layout_height="0dp"
+        android:layout_height="wrap_content"
         android:layout_marginStart="@dimen/speech_button_margin"
         android:src="@drawable/ic_speaker"
         android:visibility="gone"
         app:layout_constraintBottom_toBottomOf="@+id/keyboard_input"
+        app:layout_constraintEnd_toStartOf="@+id/action_button_container"
+        app:layout_constraintHorizontal_chainStyle="packed"
```
MattCorbett488: You shouldn't have to specify the chain style again if another link in the chain specifies it

app/src/main/res/layout/fragment_keyboard.xml
```diff
@@ -94,14 +111,82 @@
...
         android:rowCount="4"
         android:useDefaultMargins="true"
-        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintBottom_toTopOf="@+id/keyboard_bottom_bar"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toBottomOf="@+id/predictive_text" />
+        app:layout_constraintTop_toBottomOf="@+id/action_button_container" />
+
+    <LinearLayout
```
MattCorbett488: Could this be a ConstraintLayout with `app:layout_constraintHorizontal_weight` specified instead of a weighted LinearLayout?

app/src/main/res/layout/fragment_presets.xml
```diff
@@ -32,6 +40,7 @@
         android:layout_width="@dimen/categories_paging_button_width"
         android:layout_height="@dimen/categories_paging_button_height"
         android:layout_marginStart="@dimen/main_activity_side_margin"
+        android:layout_marginTop="8dp"
```
MattCorbett488: Should this be broken out into a resource since it's being used multiple times?

app/src/main/java/com/willowtree/vocable/MainActivity.kt
```diff
@@ -34,11 +29,11 @@ class MainActivity : BaseActivity() {
             .replace(R.id.fragment_container, PresetsFragment())
             .commit()
 
-        keyboard_button.action = {
-            loadFragment(KeyboardFragment())
-        }
-
         presetsViewModel = ViewModelProviders.of(this).get(PresetsViewModel::class.java)
+
+        fragment_container.addOnLayoutChangeListener { _, _, _, _, _, _, _, _, _ ->
```
MattCorbett488: Have y'all thought about migrating to ViewBinding now that 3.6 is stable, versus synthetics?

nicholas-cook: We'll look into doing this with some of our future work 👍

PR #25, opened 2020-02-26
Migrate app to use view binding

nicholas-cook:
- Migrates the app to use view binding instead of Kotlin synthetics

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -27,6 +27,9 @@ android {
     kotlinOptions {
         jvmTarget = JavaVersion.VERSION_1_8.toString()
     }
+    viewBinding {
+        enabled = true
+    }
 }
 
 dependencies {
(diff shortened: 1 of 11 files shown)


app/build.gradle
```diff
@@ -27,6 +27,9 @@ android {
     kotlinOptions {
         jvmTarget = JavaVersion.VERSION_1_8.toString()
     }
+    viewBinding {
```
MattCorbett488: beautiful

app/src/main/java/com/willowtree/vocable/keyboard/KeyboardFragment.kt
```diff
@@ -87,59 +82,59 @@ class KeyboardFragment : BaseFragment() {
...
         })
         VocableTextToSpeech.isSpeaking.observe(viewLifecycleOwner, Observer {
-            speaker_icon.isVisible = it ?: false
+            binding?.speakerIcon?.isVisible = it ?: false
         })
 
-        with(presets_button) {
-            setIconWithNoText(R.drawable.ic_presets)
-            action = {
+        with(binding?.actionButtonContainer?.presetsButton) {
```
MattCorbett488: could use ```binding?.actionButtonContainer?.presetsButton?.let {
  ...
}
```
instead so you're not having to do `this?.`?

PR #26, opened 2020-02-27
Add Room database implementation

nicholas-cook:
- Adds support for Room
- Creates a database to hold Categories and Phrases
- Add DAOs and a repository class to handle querying the database
- Hooks up the `PresetsViewModel` to use this new Room data source
- Updates fragments to use model objects instead of strings

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/presets/PresetsRepository.kt
+++ b/app/src/main/java/com/willowtree/vocable/presets/PresetsRepository.kt
@@ -1,0 +1,37 @@
+package com.willowtree.vocable.presets
+
+import android.content.Context
+import com.willowtree.vocable.room.Category
+import com.willowtree.vocable.room.Phrase
+import com.willowtree.vocable.room.VocableDatabase
+import org.koin.core.KoinComponent
+import org.koin.core.get
+
+class PresetsRepository(context: Context) : KoinComponent {
+
+    private val database = VocableDatabase.getVocableDatabase(context)
+
+    suspend fun getAllCategories(): List<Category> {
+        return database.categoryDao().getAllCategories()
+    }
+
+    suspend fun getPhrasesForCategory(categoryId: Long): List<Phrase> {
+        return database.phraseDao().getPhrasesByCategory(categoryId)
+    }
+
+    suspend fun getUserGeneratedPhrases(): List<Phrase> {
+        return database.phraseDao().getUserGeneratedPhrases()
+    }
+
+    suspend fun addPhrase(phrase: Phrase) {
+        database.phraseDao().insertPhrase(phrase)
+    }
+
+    suspend fun populateCategories(categories: List<Category>) {
+        database.categoryDao().insertCategories(*categories.toTypedArray())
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/willowtree/vocable/presets/PresetsRepository.kt
```diff
@@ -0,0 +1,37 @@
...
+import android.content.Context
+import com.willowtree.vocable.room.Category
+import com.willowtree.vocable.room.Phrase
+import com.willowtree.vocable.room.VocableDatabase
+import org.koin.core.KoinComponent
+import org.koin.core.get
+
+class PresetsRepository(context: Context) : KoinComponent {
+
+    private val database = VocableDatabase.getVocableDatabase(context)
```
dnylng: Why don't you set up a Koin module for your DB and inject it into your repository class?

app/src/main/java/com/willowtree/vocable/presets/PresetsViewModel.kt
```diff
@@ -23,7 +25,7 @@ class PresetsViewModel : ViewModel(), KoinComponent {
         const val CATEGORY_MY_SAYINGS = "My Sayings"
     }
 
-    private val sharedPrefs: VocableSharedPreferences by inject()
+    private val presetsRepository: PresetsRepository by inject()
```
dnylng: You should definitely check out using viewModel injection that Koin is offering.

app/src/main/java/com/willowtree/vocable/room/PhraseDao.kt
```diff
@@ -0,0 +1,22 @@
...
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun insertPhrase(phrase: Phrase)
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun insertPhrases(vararg phrases: Phrase)
+
+    @Delete
+    suspend fun deletePhrase(phrase: Phrase)
+
+    @Query("SELECT * FROM Phrase WHERE is_user_generated == 1")
```
dnylng: Not sure if this breaks anything but is it supposed to be `WHERE is_user_generated == 1` or `WHERE is_user_generated = 1`?

dnylng: Cuz this is based on mySQL comparisons and not kotlin/java lang right?

app/src/main/java/com/willowtree/vocable/presets/PresetsRepository.kt
nicholas-cook: 👍 We're going to look into this once we get the saving presets functionality working.

app/src/main/java/com/willowtree/vocable/room/PhraseDao.kt
nicholas-cook: It looks like Room accepts both `=` and `==` here

PR #27, opened 2020-02-28
Add ability to save presets

nicholas-cook:
- Adds new view model to handle adding new preset to the Room database
- Adds way to update category list if a custom phrase has been added

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/keyboard/KeyboardViewModel.kt
+++ b/app/src/main/java/com/willowtree/vocable/keyboard/KeyboardViewModel.kt
@@ -11,6 +11,10 @@ import org.koin.core.inject
 
 class KeyboardViewModel : BaseViewModel() {
 
+    companion object {
+        private const val PHRASE_ADDED_DELAY = 2000L
+    }
+
     private val presetsRepository: PresetsRepository by inject()
 
     private val liveShowPhraseAdded = MutableLiveData<Boolean>()
@@ -29,7 +33,7 @@ class KeyboardViewModel : BaseViewModel() {
                 )
             )
             liveShowPhraseAdded.postValue(true)
-            delay(2000)
+            delay(PHRASE_ADDED_DELAY)
             liveShowPhraseAdded.postValue(false)
         }
     }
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/willowtree/vocable/keyboard/KeyboardViewModel.kt
```diff
@@ -0,0 +1,36 @@
...
+                    System.currentTimeMillis(),
+                    System.currentTimeMillis(),
+                    true,
+                    0L,
+                    phraseStr,
+                    -1L
+                )
+            )
+            liveShowPhraseAdded.postValue(true)
+            delay(2000)
```
jequeen: We should make the time delay a variable.

app/src/main/java/com/willowtree/vocable/presets/PresetsViewModel.kt
```diff
@@ -242,12 +242,17 @@ class PresetsViewModel : BaseViewModel(), KoinComponent {
...
-            val categories = presetsRepository.getAllCategories()
+            val categories = presetsRepository.getAllCategories().toMutableList()
             if (categories.isEmpty()) {
                 populateDatabase()
             } else {
+                if (presetsRepository.getUserGeneratedPhrases().isEmpty()) {
+                    categories.removeIf {
+                        it.name == CATEGORY_MY_SAYINGS
+                    }
+                }
```
jequeen: We should only post an update if the categories list has changed.

app/src/main/java/com/willowtree/vocable/presets/PresetsViewModel.kt
```diff
@@ -298,7 +303,11 @@ class PresetsViewModel : BaseViewModel(), KoinComponent {
     fun onCategorySelected(category: Category) {
         liveSelectedCategory.postValue(category)
         backgroundScope.launch {
-            val phrases = presetsRepository.getPhrasesForCategory(category.identifier)
+            val phrases = if (category.name == CATEGORY_MY_SAYINGS) {
+                presetsRepository.getUserGeneratedPhrases()
+            } else {
+                presetsRepository.getPhrasesForCategory(category.identifier)
+            }
```
jequeen: We should support not entering duplicates into the db.

PR #28, opened 2020-02-28
Keyboard improvements

jequeen:
- Keyboard no longer type in all caps
- First letters of sentences with be capitalized
- Text new scrolls horizontally as text is typed
- Resized padding on the paging buttons


app/src/main/java/com/willowtree/vocable/keyboard/KeyboardFragment.kt
```diff
@@ -75,21 +77,40 @@ class KeyboardFragment : BaseFragment() {
...
+                    } else {
+                        binding?.keyboardInput?.append(text?.toString()?.toLowerCase(Locale.getDefault()))
+                    }
+                }
+            }
         }
     }
 
+    private fun isDefaultTextVisible(): Boolean {
+        if (binding?.keyboardInput?.text.toString() == getString(R.string.keyboard_select_letters)) {
```
nicholas-cook: We could simplify this to just:
`return binding?.keyboardInput?.text.toString() == getString(R.string.keyboard_select_letters)`

app/src/main/java/com/willowtree/vocable/keyboard/KeyboardFragment.kt
```diff
@@ -75,21 +77,40 @@ class KeyboardFragment : BaseFragment() {
...
+            with(
+                KeyboardKeyLayoutBinding.inflate(
+                    layoutInflater,
+                    binding?.keyboardKeyHolder,
+                    true
+                ).root as ActionButton
+            ) {
+                text = it.value
+                action = {
+                    val currentText = binding?.keyboardInput?.text?.toString() ?: ""
```
nicholas-cook: We should probably put a comment in here that this will mimic sentence casing.

PR #31, opened 2020-03-03
Adding portrait phrases for phones

jequeen:
- Adds portrait support for presets screen on phone.
- Created new `-land` directories to separate portrait and landscape layouts and values.

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/presets/CategoriesFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/presets/CategoriesFragment.kt
@@ -55,7 +55,7 @@ class CategoriesFragment : BaseFragment() {
                 action = {
                     viewModel.onCategorySelected(category)
                 }
-                if (!isTablet && index + 1 == maxCategories) {
+                if (!isTablet && maxCategories != 1 && index + 1 == maxCategories) {
                     layoutParams = (layoutParams as LinearLayout.LayoutParams).apply {
                         marginStart = 0
                     }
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/willowtree/vocable/presets/CategoriesFragment.kt
```diff
@@ -55,7 +55,7 @@ class CategoriesFragment : BaseFragment() {
                 action = {
                     viewModel.onCategorySelected(category)
                 }
-                if (!isTablet && index + 1 == maxCategories) {
+                if (!isTablet && maxCategories != 1 && index + 1 == maxCategories) {
```
MattCorbett488: Why 1? Worth making a constant to avoid "magic number"s?

app/src/main/res/layout-land/fragment_phrases.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<GridLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/phrases_container"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:columnCount="3"
```
MattCorbett488: Is the only thing different between these buckets (landscape, portrait, tablet) the column and row counts?

app/src/main/java/com/willowtree/vocable/presets/CategoriesFragment.kt
nicholas-cook: Yeah it was removing the start margin if there's only one category showing since the next part of the if-block is true. We can clean this up a bit to make it more clear.

app/src/main/res/layout-land/fragment_phrases.xml
MattCorbett488: That's what I'm thinking, yeah - move them into `integers.xml` and just make different buckets

nicholas-cook: We're going to handle this in an upcoming PR where we add a couple of more buckets.

PR #32, opened 2020-03-03
Landscape handset settings page

CarolineLaw:
- created tablet version, copied all the old layouts there
- changed dimens so that the handset version works
- only did landscape, didn't do portrait yet

not sure if I named them well, let me know.


app/src/main/res/values/dimens.xml
```diff
@@ -45,6 +45,13 @@
     <dimen name="settings_close_button_height">96dp</dimen>
     <dimen name="settings_close_button_width">112dp</dimen>
     <dimen name="settings_pointer_elevation">16dp</dimen>
+    
+    <dimen name="settings_handset_close_button_layout_width">72dp</dimen>
```
nicholas-cook: Instead of making new dimens with different names, you should make a new dimens file for the tablet in a directory called `values-sw600dp`. Then copy the original values into the new `sw600dp` file and update the values in this file to match the handset designs.

app/src/main/res/layout-sw600dp/activity_settings.xml
```diff
@@ -0,0 +1,155 @@
...
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginBottom="@dimen/settings_bottom_margin"
+        android:text="@string/version"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintVertical_chainStyle="packed" />
+
+    <include
```
nicholas-cook: Don't forget to update the dimens for the confirmation screen

CarolineLaw: I looked at this screen in the emulator, and it looks okay. I don't think there's anything I need to change, but let me know if there is

CarolineLaw: Screen Shot 2020-03-03 at 3 53 26 PM

PR #34, opened 2020-03-04
Tablet portrait mode for presets screen

jequeen:
- Adding support for portrait tablet on presets screen
- Moved gridlayout info to integers file

### DIFF
--- a/app/src/main/res/drawable/category_group_background.xml
+++ b/app/src/main/res/drawable/category_group_background.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <shape xmlns:android="http://schemas.android.com/apk/res/android"
     android:shape="rectangle">
-    <corners android:radius="4dp" />
+    <corners android:radius="8dp" />
     <solid android:color="@color/categoryColor" />
 </shape>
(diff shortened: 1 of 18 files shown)


app/src/main/res/drawable/category_group_background.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <shape xmlns:android="http://schemas.android.com/apk/res/android"
     android:shape="rectangle">
-    <corners android:radius="4dp" />
+    <corners android:radius="8dp" />
```
MattCorbett488: If you see this radius being used as a standard (I see it as a radius in 2 files in this PR), could be useful to extract it to a dimen so you're not updating every spot every time this changes

CarolineLaw:
so the files named `land` are landscape and everything else is portrait?

nicholas-cook:
@CarolineLaw Yep! If the directory doesn't end with `-land`, the OS will assume portrait

app/src/main/res/drawable/category_group_background.xml
jequeen: That is a fair point, I pulled those out into the `values/dimens` file. Thanks for the feedback!

PR #38, opened 2020-03-04
Adding settings portrait handset

CarolineLaw:
- added new layouts for landscape
- modified layout for portrait


app/src/main/res/layout-sw600dp/vocable_confirmation_dialog.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
```
nicholas-cook: It looks like this file is identical to `layout/vocable_confirmation_dialog`. If that's the case, you only need to keep the one in the base `layout` directory and it will get used by all form factors.

app/src/main/res/layout/activity_settings.xml
```diff
@@ -98,24 +98,25 @@
         android:layout_height="@dimen/settings_action_button_layout_height"
         android:id="@+id/privacy_policy_button"
         style="@style/ActionButton"
-        android:layout_marginStart="@dimen/settings_action_button_margin"
-        android:layout_marginEnd="@dimen/settings_margin_small"
-        android:layout_marginBottom="@dimen/settings_action_button_bottom_margin"
+        android:layout_marginStart="@dimen/settings_margin_default"
+        android:layout_marginEnd="@dimen/settings_margin_default"
+        android:layout_marginBottom="8dp"
```
nicholas-cook: This should be in a `dimen` file.

PR #29, opened 2020-03-02
Add support for landscape phone layout

nicholas-cook:
- Adds layout files with qualifiers to differentiate between phone and tablet portrait layouts
- Adds `dimen` and `bool` files to help with styling
- Portrait support for both phone and tablet will come in a future PR

PR #40, opened 2020-03-05
Clean up/bug fixes

jequeen:
- Fixed issue where when switching rotation 180 degrees facetracking controls would be inverted
- Updated version number on the settings screen
- Updated layouts to use `FragmentContainerView`

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/BaseActivity.kt
+++ b/app/src/main/java/com/willowtree/vocable/BaseActivity.kt
@@ -3,9 +3,11 @@ package com.willowtree.vocable
 import android.app.ActivityManager
 import android.content.Context
 import android.graphics.Rect
+import android.hardware.display.DisplayManager
 import android.os.Bundle
 import android.util.DisplayMetrics
 import android.util.Log
+import android.view.Surface
 import android.view.View
 import android.widget.Toast
 import androidx.annotation.CallSuper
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/willowtree/vocable/BaseActivity.kt
```diff
@@ -43,6 +46,61 @@ abstract class BaseActivity : AppCompatActivity() {
...
+    }
+
+    /**
+     * If the device rotates 180 degrees (portrait to portrait/landscape to landscape), the
+     * activity won't be destroyed and recreated. This means that the FaceTrackFragment will not
+     * reset its camera positioning. The only way to reset it currently is to create a new
+     * instance of the fragment and add it to the activity.
+     * @param tag The tag to use for the FaceTrackFragment, should be unique to the orientation
+     */
+    private fun resetFaceTrackFragment(tag: String) {
```
CarolineLaw: nice 👍

app/src/main/res/layout-land/activity_settings.xml
```diff
@@ -7,7 +7,7 @@
     android:orientation="vertical"
     android:theme="@style/SettingsTheme">
 
-    <fragment
+    <androidx.fragment.app.FragmentContainerView
```
CarolineLaw: how come we're using FragmentContainerView instead of fragment? just curious :]

nicholas-cook: This was related to the comments Matt Corbett made in #29 - `FragmentContainerView` is the new preferred method of adding fragments to an activity.

CarolineLaw: neat!

PR #60, opened 2020-03-09
Paging disable update

jequeen:
- Disables paging if there's only one page for touch
- Disables tts on the settings page for all the buttons

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/customviews/VocableImageButton.kt
+++ b/app/src/main/java/com/willowtree/vocable/customviews/VocableImageButton.kt
@@ -29,19 +29,17 @@ class VocableImageButton @JvmOverloads constructor(
     }
 
     override fun onPointerEnter() {
-        if (this.isEnabled) {
-            buttonJob = backgroundScope.launch {
-                uiScope.launch {
-                    isSelected = true
-                }
+        buttonJob = backgroundScope.launch {
+            uiScope.launch {
+                isSelected = true
+            }
 
-                delay(DEFAULT_TTS_TIMEOUT)
+            delay(DEFAULT_TTS_TIMEOUT)
 
-                uiScope.launch {
-                    isSelected = false
-                    isPressed = true
-                    performAction()
-                }
+            uiScope.launch {
+                isSelected = false
+                isPressed = true
+                performAction()
             }
         }
     }
--- a/app/src/main/java/com/willowtree/vocable/customviews/VocableImageButton.kt
+++ b/app/src/main/java/com/willowtree/vocable/customviews/VocableImageButton.kt
@@ -29,17 +29,19 @@ class VocableImageButton @JvmOverloads constructor(
     }
 
     override fun onPointerEnter() {
-        buttonJob = backgroundScope.launch {
-            uiScope.launch {
-                isSelected = true
-            }
+        if (this.isEnabled) {
+            buttonJob = backgroundScope.launch {
+                uiScope.launch {
+                    isSelected = true
+                }
 
-            delay(DEFAULT_TTS_TIMEOUT)
+                delay(DEFAULT_TTS_TIMEOUT)
 
-            uiScope.launch {
-                isSelected = false
-                isPressed = true
-                performAction()
+                uiScope.launch {
+                    isSelected = false
+                    isPressed = true
+                    performAction()
+                }
             }
         }
     }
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
```diff
@@ -303,6 +304,19 @@ class PresetsFragment : BaseFragment() {
...
+
+            if (phrasesAdapter.numPages == 1) {
+                setPagingButtonsEnabled(false)
+            } else {
+                setPagingButtonsEnabled(true)
+            }
+        }
+
+        private fun setPagingButtonsEnabled(enable: Boolean) {
+            binding.let {
```
nicholas-cook: So the convention here is to do `binding?.let`, that way you can just call `it.` without the `?` each time.

app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
```diff
@@ -303,6 +304,19 @@ class PresetsFragment : BaseFragment() {
             }
             numPages = ceil(phrases.size / maxPhrases.toDouble()).toInt()
             notifyDataSetChanged()
+
+            if (phrasesAdapter.numPages == 1) {
```
nicholas-cook: You can simplify this by just calling `setPagingButtonsEnabled(phrasesAdapter.numPages > 1)`

app/src/main/res/drawable/button_default_background.xml
```diff
@@ -2,5 +2,6 @@
 <selector xmlns:android="http://schemas.android.com/apk/res/android">
     <item android:drawable="@drawable/button_selected_background" android:state_pressed="true" />
     <item android:drawable="@drawable/button_highlighted_background" android:state_selected="true"/>
-    <item android:drawable="@drawable/button_background"/>
+    <item android:drawable="@drawable/button_background" android:state_enabled="true"/>
```
nicholas-cook: I don't think you have to specify `android:state_enabled="true"`. You can just move your enabled drawable to the bottom of the list after the disabled file. These drawable selectors act like if-blocks, so if it doesn't find a matching state, it'll just default to the last on listed.
So swap lines 5 and 6 and remove the state from the new line 6 -
`<item android:drawable="@drawable/button_background"/>`

app/src/main/res/values/colors.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <color name="colorPrimary">#3831A0</color>
+    <color name="colorPrimaryDisabled">#803831A0</color>
```
nicholas-cook: Knew it was the first two digits 😛

app/src/main/java/com/willowtree/vocable/customviews/VocableImageButton.kt
```diff
@@ -29,17 +29,19 @@ class VocableImageButton @JvmOverloads constructor(
...
-                isSelected = true
-            }
-
-            delay(DEFAULT_TTS_TIMEOUT)
-
-            uiScope.launch {
-                isSelected = false
-                isPressed = true
-                performAction()
+        if (this.isEnabled) {
```
nicholas-cook: I added a check to line 170 of `BaseActivity` so it won't recognize the view intersect if it's disabled. So you might not need this here.

app/src/main/res/drawable/button_default_background.xml
nicholas-cook: The same goes for the other two files below.

app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
```diff
@@ -303,6 +304,19 @@ class PresetsFragment : BaseFragment() {
...
+                setPagingButtonsEnabled(false)
+            } else {
+                setPagingButtonsEnabled(true)
+            }
+        }
+
+        private fun setPagingButtonsEnabled(enable: Boolean) {
+            binding.let {
+                it?.phrasesForwardButton?.isEnabled = enable
```
nicholas-cook: Do we want to disable the buttons or hide them?

CarolineLaw: they are disabled

app/src/main/java/com/willowtree/vocable/customviews/VocableImageButton.kt
CarolineLaw: so we added this because head tracking was still happening even if it was disabled

CarolineLaw: ah! we didn't have that change in, but josh rebased and got it so we removed this check 👍

PR #70, opened 2020-03-13
Edit sayings delete for real

CarolineLaw:
this PR actually gets deleting in haha

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/settings/EditPhrasesFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/settings/EditPhrasesFragment.kt
@@ -69,7 +69,9 @@ class EditPhrasesFragment: BaseFragment() {
             }
 
             phraseView.actionButtonContainer.deleteSayingsButton.action = {
-                editPhrasesViewModel.deletePhrase(phraseView.phraseEditText.tag as Phrase)
+                if (phraseView.phraseEditText.tag is Phrase) {
+                    editPhrasesViewModel.deletePhrase(phraseView.phraseEditText.tag as Phrase)
+                }
             }
 
             binding?.editPhrasesContainer?.addView(phraseView.root)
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/willowtree/vocable/settings/EditPhrasesFragment.kt
```diff
@@ -63,6 +67,11 @@ class EditPhrasesFragment: BaseFragment() {
                     }
                 }
             }
+
+            phraseView.actionButtonContainer.deleteSayingsButton.action = {
+                editPhrasesViewModel.deletePhrase(phraseView.phraseEditText.tag as Phrase)
```
nicholas-cook: We should probably wrap this in an if-block:
```
if (phraseView.phraseEditText.tag is Phrase) {
  editPhrasesViewModel.deletePhrase(phraseView.phraseEditText.tag as Phrase)
}
```

app/src/main/java/com/willowtree/vocable/settings/EditPresetsFragment.kt
```diff
@@ -14,13 +14,17 @@ import androidx.viewpager2.widget.ViewPager2
...
 import com.willowtree.vocable.databinding.FragmentEditPresetsBinding
+import com.willowtree.vocable.databinding.PhraseEditLayoutBinding
 import com.willowtree.vocable.room.Phrase
+import org.koin.android.ext.android.bind
 import java.lang.Math.ceil
 
 class EditPresetsFragment : BaseFragment() {
 
     private var binding: FragmentEditPresetsBinding? = null
+    private var bindingEditPhrases: PhraseEditLayoutBinding? = null
```
nicholas-cook: What is this second biding used for?

CarolineLaw: I think I was using it for something else, but it's not used anymore. Thank you!!!

PR #75, opened 2020-03-17
Edit mode keyboard!!!!!

CarolineLaw:
can now edit phrases!

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
@@ -116,17 +116,13 @@ class EditKeyboardFragment : BaseFragment() {
 
         binding?.keyboardInput?.setText(phrase.utterance)
 
-        binding?.keyboardClearButton?.let {
-            it.action = {
-                binding?.keyboardInput?.setText(R.string.keyboard_select_letters)
-            }
+        binding?.keyboardClearButton?.action = {
+            binding?.keyboardInput?.setText(R.string.keyboard_select_letters)
         }
 
-        binding?.keyboardSpaceButton?.let {
-            it.action = {
-                if (!isDefaultTextVisible() && binding?.keyboardInput?.text?.endsWith(' ') == false) {
-                    binding?.keyboardInput?.append(" ")
-                }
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
```diff
@@ -0,0 +1,230 @@
...
+                            viewModel.updatePhrase(phrase)
+                        }
+                    }
+                }
+            }
+        }
+
+        binding?.keyboardInput?.setText(phrase.utterance)
+
+        binding?.keyboardClearButton?.let {
```
nicholas-cook: If you're only doing one operation inside the `let` block, you can just drop the `let`. Then it's just:
`binding?.keyboardClearButton?.action = { ... }`
The same goes for other views below.

app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
```diff
@@ -0,0 +1,230 @@
...
+            it.keyboardClearButton.isEnabled = enable
+            it.keyboardInput.isEnabled = enable
+            it.keyboardKeyHolder.isEnabled = enable
+        }
+    }
+
+    private fun toggleDialogVisibility(visible: Boolean) {
+        binding?.editConfirmation?.root?.let {
+            it.isVisible = visible
+            it.post {
```
nicholas-cook: You can get rid of this `post`

app/src/main/res/layout/fragment_edit_keyboard.xml
```diff
@@ -0,0 +1,221 @@
...
+        android:orientation="horizontal"
+        android:visibility="gone"
+        android:weightSum="4"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@+id/keyboard_input">
+
+        <com.willowtree.vocable.customviews.ActionButton
+            android:layout_width="0dp"
+            android:layout_height="104dp"
```
nicholas-cook: Hardcoded values

PR #77, opened 2020-03-17
Delete confirmation dialog

CarolineLaw:
confirmation dialog appears

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/settings/EditPhrasesFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/settings/EditPhrasesFragment.kt
@@ -8,12 +8,14 @@ import android.view.ViewGroup
 import android.widget.GridLayout
 import androidx.core.view.children
 import androidx.core.view.isInvisible
+import androidx.core.view.isVisible
 import androidx.core.view.updateMargins
 import androidx.lifecycle.ViewModelProviders
 import com.willowtree.vocable.BaseFragment
 import com.willowtree.vocable.R
 import com.willowtree.vocable.customviews.PointerListener
 import com.willowtree.vocable.databinding.FragmentEditPhrasesBinding
+import com.willowtree.vocable.databinding.FragmentEditPresetsBinding
 import com.willowtree.vocable.databinding.PhraseButtonBinding
 import com.willowtree.vocable.databinding.PhraseEditLayoutBinding
 import com.willowtree.vocable.presets.PhrasesFragment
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/willowtree/vocable/settings/EditPhrasesFragment.kt
```diff
@@ -93,6 +95,42 @@ class EditPhrasesFragment: BaseFragment() {
...
+
+    private fun setSettingsButtonsEnabled(enable: Boolean) {
+        editPhrasesViewModel.setEditButtonsEnabled(enable)
+
+    }
+
+    private fun toggleDialogVisibility(visible: Boolean) {
+        binding?.deleteConfirmation?.root?.let {
+            it.isVisible = visible
+            it.post {
```
nicholas-cook: You don't need the `post` here so you should be able to get rid of the `let` too

app/src/main/java/com/willowtree/vocable/settings/EditPhrasesFragment.kt
```diff
@@ -71,7 +73,7 @@ class EditPhrasesFragment: BaseFragment() {
 
             phraseView.actionButtonContainer.deleteSayingsButton.action = {
                 if (phraseView.phraseEditText.tag is Phrase) {
-                    editPhrasesViewModel.deletePhrase(phraseView.phraseEditText.tag as Phrase)
+                    showDeletePhraseDialog(phraseView)
```
nicholas-cook: You can actually just pass the `Phrase` object directly here instead of the view. You still have access to the `phrase` from line 56. I think this means we don't actually have to set it as a tag on the view either.

app/src/main/java/com/willowtree/vocable/settings/EditPresetsFragment.kt
```diff
@@ -130,6 +130,9 @@ class EditPresetsFragment : BaseFragment() {
                     phrasesAdapter.setPhrases(phrases)
                     // Move adapter to middle so user can scroll both directions
                     val middle = phrasesAdapter.itemCount / 2
+                    if(phrasesAdapter.numPages == 0) {
```
nicholas-cook: Which cases have a `numPages` of 0?

CarolineLaw: when you delete all the phrases in My Sayings

CarolineLaw: wasn't sure how to handle this, but this works :P

nicholas-cook: Gotcha 👍

PR #80, opened 2020-03-18
Added the add feature to the edit sayings screen

CarolineLaw:
you can now add a phrase on the edit my sayings screen!

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
@@ -5,6 +5,7 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.TextView
+import androidx.core.os.bundleOf
 import androidx.core.view.children
 import androidx.core.view.isVisible
 import androidx.lifecycle.Observer
@@ -33,12 +34,8 @@ class EditKeyboardFragment : BaseFragment() {
             }
         }
 
-        fun newInstance(isEditing: Boolean): EditKeyboardFragment {
-            return EditKeyboardFragment().apply {
-                arguments = Bundle().apply {
-                    putBoolean(KEY_IS_EDITING, isEditing)
-                }
-            }
+        fun newInstance(isEditing: Boolean) = EditKeyboardFragment(). apply {
+            arguments = bundleOf(KEY_IS_EDITING to isEditing)
         }
     }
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
```diff
@@ -22,11 +22,21 @@ class EditKeyboardFragment : BaseFragment() {
...
             return EditKeyboardFragment().apply {
                 arguments = Bundle().apply {
                     putParcelable(KEY_PHRASE, phrase)
+                    putBoolean(KEY_IS_EDITING, true)
+                }
+            }
+        }
+
+        fun newInstance(isEditing: Boolean): EditKeyboardFragment {
+            return EditKeyboardFragment().apply {
```
MattCorbett488: Since y'all are using the `ktx-core` dependency, you can take advantage of [bundleOf](https://developer.android.com/reference/kotlin/androidx/core/os/package-summary#bundleOf(kotlin.Pair)) (so this would be `arguments = bundleOf(KEY_IS_EDITING to true)`

Since this is a single statement technically, you can also single line it (`fun newInstance(isEditing: Boolean) = EditKeyboardFragment().apply ....`)

app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
```diff
@@ -106,15 +117,22 @@ class EditKeyboardFragment : BaseFragment() {
...
+                            }
+                            viewModel.addNewPhrase(text.toString())
                         }
                     }
                 }
             }
         }
 
-        binding?.keyboardInput?.setText(phrase.utterance)
+        if (phrase == null) {
```
MattCorbett488: You could use the elvis operator here to condense what's inside `setText` to `phrase?.utterance ?: R.string.keyboard_select_letters`

CarolineLaw: so i get a weird error when i try to do this: 
`binding?.keyboardInput?.setText(phrase?.utterance ?: R.string.keyboard_select_letters)`

`None of the following functions can be called with the arguments supplied `

MattCorbett488: We done resolved it - had to wrap `R.string.keyboard_select_letters` with `getString()` since the `setText` being called was expecting a string

PR #82, opened 2020-03-19
Fix crash, migrate database

CarolineLaw:


app/src/main/java/com/willowtree/vocable/room/VocableDatabase.kt
```diff
@@ -4,17 +4,26 @@ import android.content.Context
...
-@Database(entities = [Category::class, Phrase::class], version = 1)
+@Database(entities = [Category::class, Phrase::class], version = 2)
 abstract class VocableDatabase : RoomDatabase() {
 
     companion object {
         private var vocableDatabase: VocableDatabase? = null
 
+        private val MIGRATION_1_2: Migration = object : Migration(1, 2) {
+            override fun migrate(database: SupportSQLiteDatabase) {
+                database.execSQL("DROP INDEX index_Phrase_utterance")
```
nicholas-cook: We should put a comment on this migration to specify what it's doing/why it's needed

CarolineLaw: `// Need to drop this index from the Phrase table to migrate`

sound okay?

nicholas-cook: I would go with something like -
`// Phrases no longer have to have unique utterances`

PR #91, opened 2020-03-24
Settings selection mode

jequeen:
- Adding selection mode screen with a head tracking button. Eventually we will add more options but for now it will only be head tracking.

- Pulled in some crash fixes from our 1.2.0 release branch


app/src/main/res/layout/selection_mode_options_layout.xml
```diff
@@ -0,0 +1,83 @@
...
+        android:layout_rowWeight="1"
+        android:layout_columnWeight="1"
+        android:layout_marginEnd="@dimen/selection_mode_button_end_margin"
+        android:layout_marginBottom="@dimen/selection_mode_buttom_bottom_margin"
+        android:background="@drawable/settings_group_background"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent">
+
+        <TextView
+            android:layout_width="wrap_content"
```
nicholas-cook: It'll probably be good to go ahead an make a style for this `TextView` and the `SwitchCompat` below since future settings options will share the same properties.

PR #104, opened 2020-03-27
Dwell time

jequeen:
- Added default dwell time to shared preferences, updated button class
- Added dwell time functionality with max and min values

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/customviews/VocableButton.kt
+++ b/app/src/main/java/com/willowtree/vocable/customviews/VocableButton.kt
@@ -10,6 +10,7 @@ import android.text.style.ImageSpan
 import android.util.AttributeSet
 import androidx.annotation.DrawableRes
 import androidx.appcompat.widget.AppCompatButton
+import com.willowtree.vocable.settings.SensitivityFragment
 import com.willowtree.vocable.utils.SpokenText
 import com.willowtree.vocable.utils.VocableSharedPreferences
 import com.willowtree.vocable.utils.VocableTextToSpeech
@@ -28,6 +29,7 @@ open class VocableButton @JvmOverloads constructor(
     defStyle: Int = 0
 ) : AppCompatButton(context, attrs, defStyle),
     PointerListener,
+    SharedPreferences.OnSharedPreferenceChangeListener,
     KoinComponent {
 
     companion object {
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/willowtree/vocable/utils/VocableSharedPreferences.kt
```diff
@@ -16,6 +16,8 @@ class VocableSharedPreferences : KoinComponent {
         const val KEY_HEAD_TRACKING_ENABLED = "KEY_HEAD_TRACKING_ENABLED"
         const val KEY_SENSITIVITY = "KEY_SENSITIVITY"
         const val DEFAULT_SENSITIVITY = 0.1F
+        const val DWELL_TIME = "DWELL_TIME"
```
nicholas-cook: We should name this `KEY_DWELL_TIME`

app/src/main/java/com/willowtree/vocable/utils/VocableSharedPreferences.kt
```diff
@@ -16,6 +16,8 @@ class VocableSharedPreferences : KoinComponent {
...
+        const val DEFAULT_DWELL_TIME = 1000L
```
nicholas-cook: We should only define these defaults in one place. Probably best to use the ones in the fragment.
`private const val DEFAULT_SENSITIVITY = SensitivityFragment.MEDIUM_SENSITIVITY`
`private const val DEFAULT_DWELL_TIME = SensitivityFragment. DWELL_TIME_ONE_SECOND`

app/src/main/java/com/willowtree/vocable/customviews/VocableButton.kt
```diff
@@ -36,7 +40,21 @@ open class VocableButton @JvmOverloads constructor(
     private val backgroundScope = CoroutineScope(Dispatchers.IO)
     private val uiScope = CoroutineScope(Dispatchers.Main)
 
+    private val sharedPrefs: VocableSharedPreferences by inject()
+    protected var dwellTime: Long
+
+    private val sharedPrefsListener =
```
MattCorbett488: Would this be better provided as a singleton in Koin? Right now it'll instantiate a listener for every button when they could probably share that functionality

nicholas-cook: What would be the best approach for this? Would the dwell time be provided by Koin and something we update with a global listener? Or would there be a custom class that overrides the base shared prefs listener that we provide with Koin? And that custom listener has something like a `getDwellTime` method.

MattCorbett488: Another option could be to make this an `object` so you're not creating a new instance for every button, so just change this to
```
object : SharedPreferences.OnSharedPreferenceChangeListener { _, key ->
            when (key) {
                VocableSharedPreferences.DWELL_TIME -> {
                    dwellTime = sharedPrefs.getDwellTime()
                }
            }
        }
```

app/src/main/res/values/strings.xml
```diff
@@ -63,6 +63,7 @@
     <string name="hover_time_text">Hover Time</string>
     <string name="cursor_sensitivity_text">Cursor Sensitivity</string>
     <string name="hover_time_amount_text">%s seconds</string>
+    <string name="hover_time_one_text">1 second</string>
```
MattCorbett488: You could potentially provide this as a [plurals](https://developer.android.com/guide/topics/resources/string-resource#Plurals) resource

nicholas-cook: Yeah this got a little weird with plurals since it's a float value. So we followed the advice in this thread: https://stackoverflow.com/a/55006692

PR #118, opened 2020-04-03
Update database and models to handle new JSON pre-population

nicholas-cook:
- Preset categories and phrases will now be populated from the `presets.json` file
- Updated database schema to handle new models and relations
- This includes a database migration that preserves user data and a new cross reference model
- Added Moshi to handle JSON parsing
- Adds new `SplashActivity` and `SplashViewModel` to handle database population on launch
**NOTE**: A future PR will handle proper locale access for categories and phrases


app/src/main/java/com/willowtree/vocable/keyboard/KeyboardViewModel.kt
```diff
@@ -16,23 +21,43 @@ class KeyboardViewModel : BaseViewModel() {
     }
 
     private val presetsRepository: PresetsRepository by inject()
+    private val mySayingsCategoryId: String =
+        get<Context>().getString(R.string.category_my_sayings_id)
```
daniel-cook, line 25: I'm not familiar with this syntax `get<Context>()`.  Is this just the App Context from your Koin module?

app/src/main/java/com/willowtree/vocable/keyboard/KeyboardViewModel.kt
```diff
@@ -16,23 +21,43 @@ class KeyboardViewModel : BaseViewModel() {
...
-                    0L,
-                    phraseStr,
-                    categoryId
+            val mySayingsCategory =
+                presetsRepository.getCategoryById(mySayingsCategoryId)
+            val phraseId = UUID.randomUUID().toString()
+            val mySayingsPhrases =
+                presetsRepository.getPhrasesForCategory(mySayingsCategory.categoryId)
+            with(presetsRepository) {
+                // TODO: Use currently set Locale
```
daniel-cook, line 38: Did you mean to do this before completing PR or is it still outstanding?

CarolineLaw: this will be addressed in another PR

app/src/main/java/com/willowtree/vocable/keyboard/KeyboardViewModel.kt
CarolineLaw: Yes I think Koin allows us to have a context in the view model

app/src/main/java/com/willowtree/vocable/room/VocableDatabase.kt
```diff
@@ -20,11 +25,52 @@ abstract class VocableDatabase : RoomDatabase() {
             }
         }
 
+        private val MIGRATION_2_3: Migration = object : Migration(2, 3) {
```
daniel-cook, line 28: Do you have the time to write unit tests around this migration?  This all looks good to me, but some tests would make me feel better.

app/src/main/java/com/willowtree/vocable/keyboard/KeyboardViewModel.kt
chrisestesaa01: Also, It might be better to have this just return the resource id and get the string in the view, rather than accessing the context in the viewmodel, even if it is being done through your DI.

app/src/main/java/com/willowtree/vocable/room/VocableDatabase.kt
chrisestesaa01: Yeah, the only room migration I have experience with is the default 'drop everything'.

app/src/main/java/com/willowtree/vocable/splash/SplashViewModel.kt
```diff
@@ -0,0 +1,159 @@
...
+import org.koin.core.inject
+import java.nio.charset.Charset
+import java.util.*
+
+class SplashViewModel : BaseViewModel() {
+
+    private val presetsRepository: PresetsRepository by inject()
+    private val moshi: Moshi by inject()
+    private val sharedPrefs: VocableSharedPreferences by inject()
+    private val numbersCategoryId = get<Context>().getString(R.string.category_123_id)
```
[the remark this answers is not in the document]
CarolineLaw: nope, the category name is 123 so the id's name is 123 haha

app/src/main/java/com/willowtree/vocable/splash/SplashViewModel.kt
```diff
@@ -0,0 +1,159 @@
...
+    private val mySayingsCategoryId = get<Context>().getString(R.string.category_my_sayings_id)
+
+    private val liveExitSplash = MutableLiveData<Boolean>()
+    val exitSplash: LiveData<Boolean> = liveExitSplash
+
+    init {
+        populateDatabase()
+    }
+
+    private fun populateDatabase() {
```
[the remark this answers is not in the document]
smithks: I agree with Daniel, this ties the database initiation closely to the splash viewmodel when ideally the splash viewmodel should just tell the repository to do it's thing and let it know when it's done so it can proceed.

app/src/main/java/com/willowtree/vocable/room/VocableDatabase.kt
nicholas-cook: Just added some tests for this migration!

app/src/main/java/com/willowtree/vocable/keyboard/KeyboardViewModel.kt
```diff
@@ -16,23 +21,43 @@ class KeyboardViewModel : BaseViewModel() {
...
-                    System.currentTimeMillis(),
-                    true,
-                    0L,
-                    phraseStr,
-                    categoryId
+            val mySayingsCategory =
+                presetsRepository.getCategoryById(mySayingsCategoryId)
+            val phraseId = UUID.randomUUID().toString()
+            val mySayingsPhrases =
+                presetsRepository.getPhrasesForCategory(mySayingsCategory.categoryId)
```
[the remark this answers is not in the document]
nicholas-cook: There's no difference, just looks like I forgot to update the reference here.

app/src/main/java/com/willowtree/vocable/room/CategoryWithPhrases.kt
```diff
@@ -1,13 +1,15 @@
...
 import androidx.room.Relation
 
 data class CategoryWithPhrases(
     @Embedded val category: Category,
     @Relation(
-        parentColumn = "identifier",
-        entityColumn = "category_id"
+        parentColumn = "category_id",
+        entityColumn = "phrase_id",
+        associateBy = Junction(CategoryPhraseCrossRef::class)
```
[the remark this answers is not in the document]
nicholas-cook: So this class isn't an entity that's stored in the database. It's just a Room data class that helps with querying the junctions between categories and phrases. We made categories and phrases have a many-to-many relationship since a category has many phrases and a phrase can be part of many categories. There's more info [here](https://developer.android.com/training/data-storage/room/relationships#many-to-many).

app/src/main/java/com/willowtree/vocable/splash/SplashViewModel.kt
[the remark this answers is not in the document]
nicholas-cook: Moved this logic into the repository!

app/src/main/java/com/willowtree/vocable/room/CategoryWithPhrases.kt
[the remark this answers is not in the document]
nicholas-cook: You can run queries on the `CategoryPhraseCrossRef`, but it'll just return single id values instead of objects. The `CategoryWithPhrases` will return full objects.

app/src/main/java/com/willowtree/vocable/keyboard/KeyboardViewModel.kt
nicholas-cook: Moved passing this in via a view model factory

app/src/main/java/com/willowtree/vocable/room/CategoryWithPhrases.kt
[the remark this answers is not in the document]
smithks: Ah very cool, thanks for the explanation

PR #120, opened 2020-04-06
Use device locale to load in correct preset values

nicholas-cook:
- Adds the `LocaleUtils` class to handle loading in the string value for the device's locale
- Updates button and text-to-speech logic to handle using our new locale-string pair for displaying and speaking phrases and categories


app/src/main/java/com/willowtree/vocable/presets/PhrasesFragment.kt
```diff
@@ -51,8 +51,8 @@ class PhrasesFragment : BaseFragment() {
             val phraseButton =
                 PhraseButtonBinding.inflate(inflater, binding?.phrasesContainer, false)
             with(phraseButton.root as VocableButton) {
-                // TODO: Use currently set Locale
-                text = phrase.localizedUtterance[Locale.US.language]
+                val pair = phrase.getLocalizedPair()
```
dnylng, line 54: You can destructure this pair to make it more readable. `val (locale, phrase) = phrase.getLocalizedPair()` so that in the set text you dont have to do `setText(pair.first, pair.second)` you just do `setText(locale, phrase)`.

app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
```diff
@@ -100,7 +100,7 @@ class EditKeyboardFragment : BaseFragment() {
         super.onViewCreated(view, savedInstanceState)
 
         binding?.backButton?.action = {
-            if (binding?.keyboardInput?.text.toString() == phrase?.localizedUtterance?.get(Locale.US.language) || arguments?.getBoolean(
+            if (binding?.keyboardInput?.text.toString() == phrase?.getLocalizedText() || arguments?.getBoolean(
```
dnylng, line 103: It's a lil hard to make our what this boolean is checking for. Additionally, maybe these conditions can be leveraged out elsewhere so that there isn't too much logic in the Fragment? Idk if a helper class or viewmodel can get the fragment the info it needs, not too much of a worry for me but something to consider.

app/src/main/java/com/willowtree/vocable/keyboard/KeyboardViewModel.kt
```diff
@@ -31,14 +31,13 @@ class KeyboardViewModel(numbersCategoryId: String, mySayingsCategoryId: String)
...
             with(presetsRepository) {
-                // TODO: Use currently set Locale
                 addPhrase(
                     Phrase(
                         phraseId,
                         System.currentTimeMillis(),
                         true,
                         System.currentTimeMillis(),
-                        mapOf(Pair(Locale.US.language, phraseStr)),
+                        mapOf(Pair(Locale.getDefault().toString(), phraseStr)),
```
dnylng, line 40: Confused on what you're passing in here. Are you converting this pair into a map? Pairs are pretty expensive so I try to avoid if I can. Would this be the same if you had kotlin extensions to do `Locale.getDefault().toString() to phaseStr`?

app/src/main/java/com/willowtree/vocable/settings/EditPhrasesFragment.kt
```diff
@@ -56,8 +56,7 @@ class EditPhrasesFragment : BaseFragment() {
             val phraseView =
                 PhraseEditLayoutBinding.inflate(inflater, binding?.editPhrasesContainer, false)
             with(phraseView.root) {
-                // TODO: Use currently set Locale
-                phrase_edit_text.text = phrase.localizedUtterance[Locale.US.language]
+                phrase_edit_text.text = phrase.getLocalizedText()
```
dnylng, line 59: Not sure if you've already re-factored some of this out but try to avoid Kotlin synthetics! I thought I saw binding somewhere before, didn't yall use view binding?

app/src/main/java/com/willowtree/vocable/utils/LocaleUtils.kt
```diff
@@ -0,0 +1,59 @@
+package com.willowtree.vocable.utils
+
+import java.util.*
+
+object LocaleUtils {
```
dnylng, line 5: Any static utils classes will need to have tests IMO. Please test it, specially since the app will be pretty reliant on this.

app/src/main/java/com/willowtree/vocable/presets/PhrasesFragment.kt
nicholas-cook: Oh nice! I didn't know this was a thing in Kotlin

app/src/main/java/com/willowtree/vocable/settings/EditPhrasesFragment.kt
nicholas-cook: Yeah we've moved the app over to binding, so this one must have slipped through. Nice catch!

app/src/main/java/com/willowtree/vocable/keyboard/KeyboardViewModel.kt
nicholas-cook: Yeah if you call `mapOf(Pair(...))`, it will convert it into a map with `pair.first` being the key and `pair.second` being the value.
I think `Locale.getDefault().toString() to phaseStr` also creates a Pair

nicholas-cook: Is it less expensive to do something like `mutableMapOf().apply { put(Locale.getDefault.toString, phraseStr) }`?

dnylng: Ahh, not sure but I doubt it. In any case, I thought the `to` thing returned a map forgot it was a pair. This is probably a case of premature optimization. I'd look into if it's possible to avoid creating pairs if possible with an alternative data structure but I think this makes it more readable at least.

app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
nicholas-cook: I pulled stuff out into variables so it's more clear 👍

PR #122, opened 2020-04-06
Change the y so that users can reach the bottom of the screen

CarolineLaw:


app/src/main/java/com/willowtree/vocable/facetracking/FaceTrackingViewModel.kt
```diff
@@ -1,15 +1,19 @@
...
 import androidx.lifecycle.ViewModel
 import com.google.ar.core.AugmentedFace
 import com.google.ar.sceneform.math.Vector3
+import com.willowtree.vocable.R
 import com.willowtree.vocable.utils.VocableSharedPreferences
 import kotlinx.coroutines.*
 import org.koin.core.KoinComponent
+import org.koin.core.get
 import org.koin.core.inject
+import java.util.*
```
MichaelWTA, line 16: That's a bunch of imports. But that's pretty typical for android, right?

app/src/main/java/com/willowtree/vocable/facetracking/FaceTrackingViewModel.kt
```diff
@@ -78,6 +82,10 @@ class FaceTrackingViewModel : ViewModel(), KoinComponent {
                         liveAdjustedVector.postValue(oldVector)
                     }
                     else -> {
+                        val isTablet = get<Context>().resources.getBoolean(R.bool.is_tablet)
+                        if (!isTablet) {
+                            y *= 2f
```
MichaelWTA, line 87: will this keep multiplying y by 2 every time we hit this method? So will y keep growing and growing?

app/src/main/java/com/willowtree/vocable/facetracking/FaceTrackingViewModel.kt
CarolineLaw: ahh i forgot to push cmd+option+L, i'll do that

app/src/main/java/com/willowtree/vocable/facetracking/FaceTrackingViewModel.kt
CarolineLaw: I don't think so, also you should pull this down and test it on your own device if u can!

app/src/main/java/com/willowtree/vocable/facetracking/FaceTrackingViewModel.kt
```diff
@@ -78,6 +82,10 @@ class FaceTrackingViewModel : ViewModel(), KoinComponent {
                         liveAdjustedVector.postValue(oldVector)
                     }
                     else -> {
+                        val isTablet = get<Context>().resources.getBoolean(R.bool.is_tablet)
```
nicholas-cook, line 85: We should grab this value in the `init` block. Reading from the resources files can be expensive, and we don't want to do it every time we loop through here.

PR #128, opened 2020-04-10
Edit categories implementation

CarolineLaw:
- add, edit, and delete categories is implemented
- I do not think we should have the add categories button in the app until we finish add/edit/delete custom phrases


app/src/main/java/com/willowtree/vocable/settings/EditCategoryOptionsFragment.kt
```diff
@@ -5,10 +5,15 @@ import android.view.LayoutInflater
...
 import androidx.core.view.isInvisible
+import androidx.core.view.isVisible
+import androidx.lifecycle.ViewModelProviders
 import com.willowtree.vocable.BaseFragment
+import com.willowtree.vocable.BaseViewModelFactory
 import com.willowtree.vocable.R
 import com.willowtree.vocable.databinding.FragmentEditCategoryOptionsBinding
 import com.willowtree.vocable.room.Category
+import kotlinx.android.synthetic.main.fragment_edit_categories.*
+import kotlinx.android.synthetic.main.vocable_confirmation_dialog.*
```
nicholas-cook, line 16: Make sure you remove these imports

PR #131, opened 2020-04-13
Design changes

CarolineLaw:
see commit messages~


app/src/main/java/com/willowtree/vocable/presets/MySayingsEmptyFragment.kt
```diff
@@ -0,0 +1,67 @@
...
+        arguments?.getBoolean(KEY_IS_SETTINGS)?.let { isSettings ->
+            when {
+                isSettings -> {
+                    binding?.star?.isVisible = true
+                    binding?.emptyMySayingsText?.setText(R.string.my_sayings_empty_settings)
+                }
+                !isSettings && resources.configuration.orientation == Configuration.ORIENTATION_LANDSCAPE && !resources.getBoolean(R.bool.is_tablet) -> {
+                    val param = binding?.emptyMySayingsText?.layoutParams as ConstraintLayout.LayoutParams
+                    param.setMargins(0,0,0,0)
+                    binding?.emptyMySayingsText?.layoutParams = param
```
nicholas-cook, line 48: Why this logic instead of just having a separate layout file?

app/src/main/java/com/willowtree/vocable/presets/PresetsRepository.kt
```diff
@@ -159,34 +159,28 @@ class PresetsRepository(context: Context) : KoinComponent {
...
-            sharedPrefs.setMySayings(emptySet())
-        } else {
-            val phrases = getPhrasesForCategory(mySayingsCategoryId)
-            if (phrases.isEmpty()) {
-                mySayingsCategory.hidden = true
-            }
+            )
         }
 
+        sharedPrefs.setMySayings(emptySet())
```
nicholas-cook, line 182: So I updated the places where we were setting the My Sayings category to hidden in the `edit-categories-implementation` branch. Might want to make sure there's no conflicts with that.

app/src/main/java/com/willowtree/vocable/settings/EditPhrasesViewModel.kt
```diff
@@ -54,12 +54,6 @@ class EditPhrasesViewModel(numbersCategoryId: String, mySayingsCategoryId: Strin
                         phrase.phraseId
                     )
                 )
-                val catPhraseList = getPhrasesForCategory(mySayingsCategory.categoryId)
-                if (catPhraseList.isEmpty()) {
-                    updateCategory(mySayingsCategory.apply {
-                        hidden = true
```
nicholas-cook, line 60: Same note here about the other branch

app/src/main/java/com/willowtree/vocable/presets/MySayingsEmptyFragment.kt
CarolineLaw: because if it is coming from settings it does need a margin. Is there a way for me to make two separate landscape layouts? that'd be dope :D

nicholas-cook: Ok gotcha. This is cool then, not sure if there's an easy way to get two separate layouts

PR #137, opened 2020-04-15
Design changes for release

CarolineLaw:
see attached ticket

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
@@ -124,14 +124,11 @@ class EditKeyboardFragment : BaseFragment() {
         binding?.backButton?.action = {
             val isEditing = arguments?.getBoolean(KEY_IS_EDITING) ?: false
             val textChanged = binding?.keyboardInput?.text.toString() != phrase?.getLocalizedText()
-            if (isCategory) {
+            val categoryTextChanged = binding?.keyboardInput?.text.toString() != category?.getLocalizedText()
+            if (isCategory && !categoryTextChanged || isDefaultTextVisible()) {
                 parentFragmentManager.popBackStack()
-            } else if (isEditing && !textChanged) {
-                parentFragmentManager
-                    .beginTransaction()
-                    .replace(R.id.settings_fragment_container, EditPresetsFragment())
-                    .addToBackStack(null)
-                    .commit()
+            } else if (isEditing && !textChanged || isDefaultTextVisible()) {
+                parentFragmentManager.popBackStack()
             } else {
                 showConfirmationDialog()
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
```diff
@@ -250,11 +270,15 @@ class EditKeyboardFragment : BaseFragment() {
...
             it.action = {
-                parentFragmentManager
-                    .beginTransaction()
-                    .replace(R.id.settings_fragment_container, EditPresetsFragment())
-                    .addToBackStack(null)
-                    .commit()
+                if (isCategory) {
+                    parentFragmentManager.popBackStack()
+                } else {
+                    parentFragmentManager
```
nicholas-cook, line 276: Should this one also be `parentFragmentManager.popBackStack()`?

app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
```diff
@@ -124,14 +126,11 @@ class EditKeyboardFragment : BaseFragment() {
         binding?.backButton?.action = {
             val isEditing = arguments?.getBoolean(KEY_IS_EDITING) ?: false
             val textChanged = binding?.keyboardInput?.text.toString() != phrase?.getLocalizedText()
-            if (isCategory) {
+            val categoryTextChanged = binding?.keyboardInput?.text.toString() != category?.getLocalizedText()
```
nicholas-cook, line 129: Are you changing `keyboard_input` to an `AppCompatTextView` here or in another PR?

CarolineLaw: another PR :]

app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
CarolineLaw: yes i think so, thank you!

app/src/main/java/com/willowtree/vocable/settings/EditKeyboardFragment.kt
```diff
@@ -273,11 +273,7 @@ class EditKeyboardFragment : BaseFragment() {
                 if (isCategory) {
                     parentFragmentManager.popBackStack()
                 } else {
-                    parentFragmentManager
-                        .beginTransaction()
-                        .replace(R.id.settings_fragment_container, EditPresetsFragment())
-                        .addToBackStack(null)
-                        .commit()
+                    parentFragmentManager.popBackStack()
```
nicholas-cook, line 276: You should be able to get rid of the if-statement now if they're both doing the same thing

PR #138, opened 2020-04-15
Changes for release

CarolineLaw:
disabled my sayings edit, removed reset app, changed edittext to textview

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/settings/EditCategoriesListFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/settings/EditCategoriesListFragment.kt
@@ -161,6 +161,11 @@ class EditCategoriesListFragment : BaseFragment() {
             moveCategoryDownButton.action = {
                 editCategoriesViewModel.moveCategoryDown(category)
             }
+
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/willowtree/vocable/settings/EditCategoriesListFragment.kt
```diff
@@ -161,6 +161,11 @@ class EditCategoriesListFragment : BaseFragment() {
             moveCategoryDownButton.action = {
                 editCategoriesViewModel.moveCategoryDown(category)
             }
+
+            if (category?.categoryId == getString(R.string.category_my_sayings_id)) {
+                editCategorySelectButton.isEnabled = false
```
nicholas-cook, line 166: Do we want to disable it or make it invisible?

CarolineLaw: Steve says we should disable it 🤷

CarolineLaw: i think we should make it invisible tho, but maybe we can talk about it

PR #147, opened 2020-04-20
Fix dialog

CarolineLaw:
- getting rid of the discard dialog when adding a new phrase
- nick fixed the adapter so that it update correctly


app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
```diff
@@ -275,24 +274,24 @@ class PresetsFragment : BaseFragment() {
...
-        var numPages = 0
+        override fun setItems(items: List<Any>) {
+            super.setItems(items)
 
-        fun setCategories(categories: List<Category>) {
-            with(this.categories) {
+            with(categories) {
                 clear()
-                addAll(categories)
+                items.forEach {
```
MattCorbett488, line 286: Think you can condense this down to `addAll( items.filter { it is Category } )`

app/src/main/java/com/willowtree/vocable/settings/EditPresetsFragment.kt
```diff
@@ -191,37 +190,34 @@ class EditPresetsFragment : BaseFragment() {
...
+
+            with(phrases) {
                 clear()
-                addAll(phrases)
-            }
-            numPages = if (phrases.isEmpty()) {
-                1
-            } else {
-                ceil(phrases.size / maxPhrases.toDouble()).toInt()
+                items.forEach {
```
MattCorbett488, line 202: with the calls to super, are you not basically doing this twice?

app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
MattCorbett488: Also, do you expect some of these types to _not_ be `Category`?

nicholas-cook: So `items` could be either a list of `Category` or a list of `Phrase`. So I was having trouble casting the `items` list to the right type in each of the subclasses. Would making it generic work for that? So I could have a `VocableFragmentStateAdapter<Category>` and `setItems` would be `setItems(items: List<Category>)`?

MattCorbett488: It should yeah

You could do `VocableFragmentStateAdapter<T>(fm: FragmentManager, lifecycle: Lifecycle)`
and then change `setItems` to `open fun setItems(items: List<T>)`

nicholas-cook: Cool, just updated the adapters

app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
```diff
@@ -275,78 +274,48 @@ class PresetsFragment : BaseFragment() {
...
-            notifyDataSetChanged()
+        VocableFragmentStateAdapter<Phrase>(fm, viewLifecycleOwner.lifecycle) {
 
+        override fun setItems(items: List<Phrase>) {
+            super.setItems(items)
             setPagingButtonsEnabled(phrasesAdapter.numPages > 1)
         }
 
         private fun setPagingButtonsEnabled(enable: Boolean) {
             binding?.let {
```
jstubing, line 301: You can alternatively do a binding?.apply here and then you wouldn't need to use "it" in the body

app/src/main/java/com/willowtree/vocable/settings/EditCategoriesFragment.kt
```diff
@@ -208,6 +195,7 @@ class EditCategoriesFragment : BaseFragment() {
         binding?.let {
```
jstubing, line 195: Same deal here. Using apply can clean this up a bit.

app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
nicholas-cook: Good point, we'll address these blocks going forward. There's multiple places we use this style.

PR #150, opened 2020-04-21
Increase button size

CarolineLaw:
- get rid of small thing in german keyboard
- allows head tracking over larger area
- increased size of button

### DIFF
--- a/app/src/main/res/values-de-land/arrays.xml
+++ b/app/src/main/res/values-de-land/arrays.xml
@@ -24,7 +24,6 @@
         <item>Ö</item>
         <item>Ä</item>
         <item>Y</item>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values-de-land/arrays.xml
```diff
@@ -24,7 +24,6 @@
         <item>Ö</item>
         <item>Ä</item>
         <item>Y</item>
-        <item>\'</item>
```
nicholas-cook, line 27: Does this need to be removed from the other German files?

jstubing: It looks like this was a duplicate character in the values-de-land file only.

PR #154, opened 2020-04-22
Restore selected category position on config change

jstubing:
Upon config change, this will find the page closest to the middle in the category ViewPager which contains the currently selected category. This was a bit more involved than I thought. Please review carefully and if anyone has an alternative solution, I'm interested to hear it!

Also extracted category and phrase handling to their own functions so that subscribeToViewModel isn't so bloated.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -51,6 +51,7 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     implementation 'androidx.appcompat:appcompat:1.1.0'
     implementation 'androidx.core:core-ktx:1.2.0'
+    implementation 'androidx.lifecycle:lifecycle-livedata-ktx:2.2.0'
     implementation 'androidx.constraintlayout:constraintlayout:1.1.3'
     implementation 'androidx.fragment:fragment:1.2.3'
     // Provides ARCore Session and related resources.
--- a/app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
@@ -2,6 +2,7 @@ package com.willowtree.vocable.presets
 
 import android.content.Intent
 import android.os.Bundle
+import android.util.Log
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
@@ -29,6 +30,10 @@ import kotlin.math.min
 
 class PresetsFragment : BaseFragment() {
 
+    companion object {
+        private const val KEY_CATEGORY_POSITION = "KEY_CATEGORY_POSITION"
+    }
+
     private var binding: FragmentPresetsBinding? = null
     private val allViews = mutableListOf<View>()
 
@@ -38,6 +43,8 @@ class PresetsFragment : BaseFragment() {
     private lateinit var presetsViewModel: PresetsViewModel
     private lateinit var categoriesAdapter: CategoriesPagerAdapter
     private lateinit var phrasesAdapter: PhrasesPagerAdapter
+
+    private var categoryPosition: Int? = 0
 
     override fun onCreateView(
         inflater: LayoutInflater,
@@ -169,6 +176,8 @@ class PresetsFragment : BaseFragment() {
             }
         })
 
+        categoryPosition = savedInstanceState?.getInt(KEY_CATEGORY_POSITION)
+
         SpokenText.postValue(null)
 
         presetsViewModel =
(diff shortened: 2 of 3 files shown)


app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
```diff
@@ -273,20 +228,69 @@ class PresetsFragment : BaseFragment() {
...
+
+            if (targetPosition % categoriesAdapter.numPages != 0) {
+                targetPosition %= categoriesAdapter.numPages
+            }
+
+            presetsViewModel.selectedCategory.value?.let { selectedCategory ->
+                for (i in targetPosition until targetPosition + categoriesAdapter.numPages) {
+                    val pageCategories = categoriesAdapter.getItemsByPosition(i)
+
+                    if (pageCategories.find { pageCategory -> pageCategory.categoryId == selectedCategory.categoryId } != null) {
```
nicholas-cook, line 250: I think you can write this as:
```
pageCategories.firstOrNull { it.categoryId == selectedCategory.categoryId }?.let {
  targetPosition = i
  break
}
```

app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
```diff
@@ -273,20 +228,69 @@ class PresetsFragment : BaseFragment() {
         }
     }
 
+    private fun handleCategories(categories: List<Category>) {
+        with(binding?.categoryView) {
```
nicholas-cook, line 232: I think this is one of the places where we can get rid of the `this?.` by using `binding?.categoryView?.apply { }`

app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
```diff
@@ -273,20 +228,69 @@ class PresetsFragment : BaseFragment() {
...
+                    }
+                }
+            }
+
+            this?.setCurrentItem(targetPosition, false)
+        }
+    }
+
+    private fun handlePhrases(phrases: List<Phrase>) {
+        with(binding?.phrasesView) {
```
nicholas-cook, line 262: Same comment about `this?.`

app/src/main/java/com/willowtree/vocable/utils/VocableFragmentStateAdapter.kt
```diff
@@ -54,5 +55,20 @@ abstract class VocableFragmentStateAdapter<T>(fm: FragmentManager, lifecycle: Li
         return position in 0 until itemCount
     }
 
+    /**
+     * Returns the list of items held by the page at the provided position.
+     */
+    fun getItemsByPosition(position: Int): List<T> {
```
nicholas-cook, line 61: Nice!

app/build.gradle
```diff
@@ -51,6 +51,7 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     implementation 'androidx.appcompat:appcompat:1.1.0'
     implementation 'androidx.core:core-ktx:1.2.0'
+    implementation 'androidx.lifecycle:lifecycle-livedata-ktx:2.2.0'
```
nicholas-cook, line 54: What does this plugin do?

jstubing: At the very least, it adds an observe extension on LiveData which allows you to use function references e.g. `categoryList.observe(viewLifecycleOwner, ::handleCategories)`

app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
jstubing: hmm good suggestion but the break doesn't work in there. Wonder if there's a way to do it with return@label

app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
jstubing: Good call. I'm wondering if there's a better way to handle binding such that it's a non-nullable type. I feel like it's complicating the code a lot by being nullable. For now I'll update those to use apply.

nicholas-cook: Yeah the docs say that fragment bindings are nullable because the fragment can exist in memory even after its views are destroyed. But I wonder if there's a way around it

app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
nicholas-cook: Yeah I've noticed that breaking from loops in Kotlin can be weird. I had to move a `return@` to the top of the loop for it to work in `LocaleUtils`

PR #165, opened 2020-04-23
Settings screen layout updates

jstubing:
Tweaks to allow longer titles and to prevent clipping with other views


app/src/main/res/layout/fragment_settings.xml
```diff
@@ -7,27 +7,42 @@
     android:orientation="vertical"
     android:theme="@style/SettingsTheme">
 
+    <androidx.constraintlayout.widget.Guideline
```
MattCorbett488, line 10: This worth extracting as an `<include>` since it's getting used in multiple spots?

jstubing: Good call, I updated to use an include.

PR #188, opened 2020-04-28
Fix crash when drilling down into a category

jstubing:
Fragments should use viewLifecycleOwner rather than the activity when observing LiveData, otherwise there's the potential for LiveData to be observed after the view has been destroyed.


CarolineLaw:
@nicholas-cook  so it looks like hiding the category works, and since the other features are being hidden right now i think this fix is okay for now

jstubing:
Yes, I temporarily re-enabled the add category button and everything appears to be working fine. 👍

PR #195, opened 2020-04-30
Espresso automation setup

nerdsosky:
This PR sets up the CircleCI pipeline to run the UI Tests. Currently it is setup that for PR's the test job will run which builds the app and runs the tests against it. Merges to Develop and Release branches should run the build-and-deploy step which builds the app and then deploys to AppCenter. The UI Tests will run against these builds as well but this can be disabled if it adds too much time


.circleci/config.yml
```diff
@@ -35,8 +35,82 @@ jobs:
...
+            sdkmanager --list
+
+      - save_cache:
+          key: android=tools-v1-{{ checksum "install-android-tools.sh" }}-{{ arch }}
+          paths:
+            - /Users/distiller/android-tools
+
+      - run:
+          name: create AVD
+          command: echo no | avdmanager create avd -d "pixel_xl" -n test -k "system-images;android-29;google_apis;x86_64" --force
```
BarryBryant, line 77: Will this get cached? Was curious what the typical set up is for this on automation, seems like it's a big download to run for every time we run UI tests.

Issue #194, opened 2020-04-28
Tracking cursors persists on settings screen after disabling head tracking


nerdsosky:
The orange cursor can still be seen on the settings screen after disabling head tracking selection

Steps:
Launch application
Open the settings
Disable head tracking
Navigate back to settings

Result: Orange tracking cursor is still displayed in the top left of the screen

Expected: You should not see the cursor if head tracking is disabled

![trackerPersists](https://user-images.githubusercontent.com/10504463/80535610-154d8b00-896f-11ea-9f1a-12de2568b1ae.png)

kennethWTA:
Not sure if you want a separate issue, but there's a ~1 second delay to when entering the Settings screen as well as when exiting it.
App ver.: 1.3.1
Device: Pixel 3, Android 10.

Issue #193, opened 2020-04-28
Head tracking orange dot is not visible if toggled off and on in the settings and returning to the phrase screen


nerdsosky:
If the user disables head tracking from the settings and then goes back in and re-enables it in the same session, upon returning to the keyboard screen or phrase screen head tracking is not enabled.

Steps:

1. Launch the application
2. Select the Settings button
3. Select the selection mode button
4. Turn off head tracking
5. Navigate back to the settings screen
6. Close the Settings screen
7. Open the settings screen
8. Open Selection mode
9. Turn on head tracking
10. Go back to settings
11. Exit settings to presets screen

Note: Looking away from the phone the "Move closer to the screen" alert will appear and then when you look back at the device the alert will go away and the orange dot will reappear.

Result: Orange tracking dot is not visible
Expected: Orange dot is visible when returning to preset screen

Build: 1.3.1

Device: Samsung S7 Android 8.0.0

PR #198, opened 2020-05-01
Fix head tracking

CarolineLaw:
- head tracking was enabled with the buildconfig, but it should also look at the user settings
- this pr should fix both tickets #193 and #194

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/MainActivity.kt
+++ b/app/src/main/java/com/willowtree/vocable/MainActivity.kt
@@ -28,8 +28,8 @@ class MainActivity : BaseActivity() {
         binding = ActivityMainBinding.inflate(layoutInflater)
         setContentView(binding.root)
 
-        if (BuildConfig.USE_HEAD_TRACKING) {
-            binding.pointerView.isVisible = true
+        if (!BuildConfig.USE_HEAD_TRACKING) {
+            binding.pointerView.isVisible = false
         } else {
             binding.pointerView.isVisible = sharedPrefs.getHeadTrackingEnabled()
         }
@@ -55,8 +55,8 @@ class MainActivity : BaseActivity() {
 
     override fun onResume() {
         super.onResume()
-        if (BuildConfig.USE_HEAD_TRACKING) {
-            binding.pointerView.isVisible = true
+        if (!BuildConfig.USE_HEAD_TRACKING) {
+            binding.pointerView.isVisible = false
         } else {
             binding.pointerView.isVisible = sharedPrefs.getHeadTrackingEnabled()
         }
--- a/app/src/main/java/com/willowtree/vocable/MainActivity.kt
+++ b/app/src/main/java/com/willowtree/vocable/MainActivity.kt
@@ -27,7 +27,12 @@ class MainActivity : BaseActivity() {
 
         binding = ActivityMainBinding.inflate(layoutInflater)
         setContentView(binding.root)
-        binding.pointerView.isVisible = BuildConfig.USE_HEAD_TRACKING && sharedPrefs.getHeadTrackingEnabled()
+
+        if (BuildConfig.USE_HEAD_TRACKING) {
+            binding.pointerView.isVisible = true
+        } else {
+            binding.pointerView.isVisible = sharedPrefs.getHeadTrackingEnabled()
+        }
 
         supportActionBar?.hide()
         VocableTextToSpeech.initialize(this)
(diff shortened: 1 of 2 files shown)


CarolineLaw:
thinking about this a little more, i'm not sure how we wanna handle turning head tracking on and off for testing..

app/src/main/java/com/willowtree/vocable/MainActivity.kt
```diff
@@ -10,21 +10,24 @@ import com.willowtree.vocable.customviews.PointerListener
...
+    private val sharedPrefs: VocableSharedPreferences by inject()
     private val allViews = mutableListOf<View>()
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
         binding = ActivityMainBinding.inflate(layoutInflater)
         setContentView(binding.root)
-        binding.pointerView.isVisible = BuildConfig.USE_HEAD_TRACKING
+        binding.pointerView.isVisible = BuildConfig.USE_HEAD_TRACKING && sharedPrefs.getHeadTrackingEnabled()
```
CarolineLaw, line 30: how do we wanna handle this? maybe 
```suggestion
        if (BuildConfig.USE_HEAD_TRACKING) {
            binding.pointerView.isVisible = true
        } else {
            binding.pointerView.isVisible = sharedPrefs.getHeadTrackingEnabled()
        }
```

BarryBryant: Hmmm what's the use case for the config vs shared preference?

CarolineLaw: so I think the config is for testing and the shared prefs one is the one that the user sets

BarryBryant: If there's no instance where we would want to be testing the user preferences then the suggestion makes perfect sense to me. I'm guessing we wouldn't be automating head tracking stuff, so this is really just a flag to turn it off?

CarolineLaw: i think so, but let's check with @nerdsosky

nerdsosky: Yes this flag disables head tracking for the automation tests. It was added before I was on the project but that is what we are currently using it for at the least.

app/src/main/java/com/willowtree/vocable/MainActivity.kt
```diff
@@ -27,7 +27,12 @@ class MainActivity : BaseActivity() {
 
         binding = ActivityMainBinding.inflate(layoutInflater)
         setContentView(binding.root)
-        binding.pointerView.isVisible = BuildConfig.USE_HEAD_TRACKING && sharedPrefs.getHeadTrackingEnabled()
+
+        if (BuildConfig.USE_HEAD_TRACKING) {
+            binding.pointerView.isVisible = true
```
nicholas-cook, line 32: I think this value actually defaults to `true` in the `build.gradle` file. So if it's not explicitly set to false, users won't be able to turn head tracking off.

CarolineLaw: ah i see, so should we do the other way around?

nicholas-cook: Yeah probably best to check:
``` kotlin
if (!BuildConfig.USE_HEAD_TRACKING) {
  binding.pointerView.isVisible = false
} else {
  binding.pointerView.isVisible = sharedPrefs.getHeadTrackingEnabled()
}
```

app/src/main/java/com/willowtree/vocable/MainActivity.kt
```diff
@@ -10,21 +10,29 @@ import com.willowtree.vocable.customviews.PointerListener
...
     private val allViews = mutableListOf<View>()
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
         binding = ActivityMainBinding.inflate(layoutInflater)
         setContentView(binding.root)
-        binding.pointerView.isVisible = BuildConfig.USE_HEAD_TRACKING
+
+        if (!BuildConfig.USE_HEAD_TRACKING) {
```
nicholas-cook, line 31: You shouldn't need to do this in both `onCreate` and `onResume` - both will get called the first time the activity runs, and `onResume` gets called every time a user returns to the activity

CarolineLaw: right, I found that we did need to call it in `onResume` _because_ it gets called every time a user returns to the activity. The pointer disappears again when you return to the activity I think

nicholas-cook: Do we need it in both `onCreate` and `onResume` or can we get away with it just being in `onResume`?

CarolineLaw: I'll test and check :]

CarolineLaw: I think we can get away with just having it in `onResume`, thank you!

PR #202, opened 2020-05-02
Feature/barry refactor

BarryBryant:
Big slice of this bad boi is just the resource files from crowdin 

Update to move away from manually handling localization for presets and deferring to Android system.
* Adds a resourceID to phrase & category so they can support both user generated & presets
* Respective DB migrations
* utilize an enum to contain our preset categories & keys
* Use user readable keys for preset values


app/build.gradle
```diff
@@ -89,3 +89,12 @@ dependencies {
     androidTestImplementation 'androidx.test.ext:junit:1.1.1'
     androidTestImplementation 'androidx.test:core:1.2.0'
 }
+
+// This task will clone the vocable-presets repo and copy the presets.json and presets.xml files
```
BarryBryant, line 93: remove this

app/src/main/java/com/willowtree/vocable/presets/PresetCategories.kt
```diff
@@ -0,0 +1,38 @@
...
+
+
+    fun getArrayId(): Int {
+        return when (this) {
+            GENERAL -> R.array.category_general
+            BASIC_NEEDS -> R.array.category_basic_needs
+            CONVERSATION -> R.array.category_conversation
+            ENVIRONMENT -> R.array.category_environment
+            PERSONAL_CARE -> R.array.category_personal_care
+            USER_KEYPAD -> -1 // Not localized with same convention
```
BarryBryant, line 22: Is the keypad static or could it vary by language? Could potentially move these to use the same format as the rest of the presets

app/src/main/java/com/willowtree/vocable/room/VocableDatabaseMigrations.kt
```diff
@@ -53,4 +54,59 @@ object VocableDatabaseMigrations {
...
+                phraseCursor.close()
+            }
+
+            // If we didn't pick up any sayings from scheme 3, check if there were some in
+            // scheme 2
+            val nonLocalizedSayings = VocableSharedPreferences().getMySayings()
+            if (myLocalizedSayings.isEmpty() && nonLocalizedSayings.isNotEmpty()) {
+                nonLocalizedSayings.forEach {
+                    val map = HashMap<String, String>()
+                    map["en"] = it
```
BarryBryant, line 93: does this need to be en_US

app/src/main/java/com/willowtree/vocable/presets/PresetCategories.kt
CarolineLaw: I'm pretty sure it's static, the only thing that should change in the keypad is "yes" and "no"

BarryBryant: Ok perfect, I can move this to support the new pattern then. Thanks!

app/src/main/java/com/willowtree/vocable/room/VocableDatabaseMigrations.kt
```diff
@@ -53,4 +54,59 @@ object VocableDatabaseMigrations {
...
+                while (phraseCursor.moveToNext()) {
+                    val saying = phraseCursor.getString(phraseCursor.getColumnIndex("localized_utterance"))
+                    myLocalizedSayings.add(saying)
+                }
+                phraseCursor.close()
+            }
+
+            // If we didn't pick up any sayings from scheme 3, check if there were some in
+            // scheme 2
+            val nonLocalizedSayings = VocableSharedPreferences().getMySayings()
```
nicholas-cook, line 89: Do we need this check here? I think that if the user is being migrated from 2 to 4, Room will run 2 to 3 then 3 to 4

app/src/main/java/com/willowtree/vocable/utils/LocalizedResourceUtility.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.willowtree.vocable.room.Phrase
+import org.koin.core.KoinComponent
+import org.koin.core.get
+import java.util.*
+
+class LocalizedResourceUtility(context: Context) : KoinComponent {
+
+    val resources: Resources = get<Context>().resources
+
+    fun getValueLocalePairFromCategory(category: Category): Pair<String, Locale>? {
```
nicholas-cook, line 15: Would it be worth having these methods return non-nullable values? I noticed in a few places we're defaulting to the empty string if they return null. Could we do that in the method?

app/src/main/java/com/willowtree/vocable/presets/PresetsRepository.kt
```diff
@@ -82,65 +82,48 @@ class PresetsRepository(context: Context) : KoinComponent {
         return database.categoryDao().getCategoryById(categoryId)
     }
 
-    suspend fun populateDatabase(numbersCategoryId: String, mySayingsCategoryId: String) {
+    suspend fun populateDatabase() {
```
nicholas-cook, line 85: We may want to do some kind of check here to see if the presets updated. Otherwise users will be stuck on older versions.

app/src/main/java/com/willowtree/vocable/room/VocableDatabaseMigrations.kt
BarryBryant: I don't think the actual presets repo initialization will happen before 3-4 runs so we would only wind up with the old scheme in mySayings

app/src/main/java/com/willowtree/vocable/utils/LocalizedResourceUtility.kt
BarryBryant: I like that!

app/src/main/java/com/willowtree/vocable/presets/PresetsRepository.kt
BarryBryant: Hmm good call

app/src/main/java/com/willowtree/vocable/BaseViewModelFactory.kt
```diff
@@ -9,31 +9,28 @@ import com.willowtree.vocable.settings.EditPhrasesViewModel
...
-    private val mySayingsCategoryId: String
-) : ViewModelProvider.Factory {
+class BaseViewModelFactory : ViewModelProvider.Factory {
 
     @Suppress("UNCHECKED_CAST")
     override fun <T : ViewModel?> create(modelClass: Class<T>): T {
         when {
             modelClass.isAssignableFrom(SplashViewModel::class.java) -> {
-                return SplashViewModel(numbersCategoryId, mySayingsCategoryId) as T
+                return SplashViewModel() as T
```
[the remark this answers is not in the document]
BarryBryant: Hmm might be worth leaving for the case, since it's already in place. I'm open 100% though what do you think?

app/src/main/java/com/willowtree/vocable/presets/PresetsRepository.kt
BarryBryant: I think I might leave this as a separate task, handling preset scheme/version. To avoid bloating dis wan.

app/src/main/java/com/willowtree/vocable/room/VocableDatabaseMigrations.kt
```diff
@@ -53,4 +54,59 @@ object VocableDatabaseMigrations {
             database.execSQL("ALTER TABLE Phrase_New RENAME TO Phrase")
         }
     }
+
+    val MIGRATION_3_4: Migration = object : Migration(3, 4) {
```
nicholas-cook, line 58: Don't forget to add a test for this migration

app/src/androidTest/java/com/willowtree/vocable/room/MigrationTest.kt
```diff
@@ -61,4 +64,67 @@ class MigrationTest {
...
+
+        val phraseTwoId = UUID.randomUUID().toString()
+        val creationDateTwo = System.currentTimeMillis()
+
+        helper.createDatabase(TEST_DB, 3).apply {
+            // Create mock My Sayings category
+            execSQL("INSERT INTO Category (category_id, creation_date, is_user_generated, localized_name, hidden, sort_order) VALUES ('${PresetCategories.USER_FAVORITES.id}', ${System.currentTimeMillis()}, 0, '$mySayingsCategoryNameV3', 0, ${PresetCategories.USER_FAVORITES.initialSortOrder})")
+
+            // Add custom phrases to My Sayings category
+            execSQL("INSERT INTO Phrase (phrase_id, creation_date, is_user_generated, last_spoken_date, localized_utterance, sort_order) VALUES ('$phraseOneId', $creationDateOne, 1, $creationDateOne, '$mySayingTestPhraseOneV3', 0)")
```
nicholas-cook, line 95: It might be worth inserting a couple of phrases into shared prefs to test the part of the migration where there might be some leftover values from the 2 to 3 conversion.

BarryBryant: good idea!

app/src/androidTest/java/com/willowtree/vocable/room/MigrationTest.kt
```diff
@@ -61,4 +64,117 @@ class MigrationTest {
...
+                }
+                Assert.assertTrue(myLocalizedSayings.contains(mySayingTestPhraseOneV3))
+                Assert.assertTrue(myLocalizedSayings.contains(mySayingTestPhraseTwoV3))
+                close()
+            }
+    }
+
+    @Test
+    @Throws(IOException::class)
+    fun migrate3to4_with_legacySayings() {
```
nicholas-cook, line 133: This should probably use camel case instead of snake case 🐫

BarryBryant: lol what about this hybrid

PR #206, opened 2020-05-05
Feature/initial ui tests

nerdsosky:


app/src/androidTest/java/com/willowtree/vocable/tests/MainScreenTest.kt
```diff
@@ -0,0 +1,29 @@
...
+        mainScreen.apply {
+            firstPhrase.tap()
+            currentText.assertTextMatches("Please")
+        }
+    }
+
+    @Test
+    fun verifyDefaultCategoriesExist() {
+        mainScreen.apply {
+            mainScreen.verifyDefaultCategories()
```
stechkev, line 26: I'd probs also call this verifyDefaultCategoriesExist() so that if it fails in there we still get an idea of what we're trying to verify in the ci logs.

nerdsosky: updated

PR #231, opened 2020-05-15
clean up build flag, fix boolean error, improve readability

BarryBryant:
Description: 
* Had a boolean reversed so was always escaping and not setting up head tracking on release

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/BaseActivity.kt
+++ b/app/src/main/java/com/willowtree/vocable/BaseActivity.kt
@@ -42,9 +42,9 @@ abstract class BaseActivity : AppCompatActivity() {
         super.onCreate(savedInstanceState)
 
         val shouldForceDisableHeadTracking = !BuildConfig.USE_HEAD_TRACKING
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/willowtree/vocable/BaseActivity.kt
```diff
@@ -40,7 +40,11 @@ abstract class BaseActivity : AppCompatActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        if (BuildConfig.USE_HEAD_TRACKING && !checkIsSupportedDeviceOrFinish()) {
+
+        val shouldForceDisableHeadTracking = !BuildConfig.USE_HEAD_TRACKING
+        val isSupportedDevice = checkIsSupportedDeviceOrFinish()
+
+        if (shouldForceDisableHeadTracking || !isSupportedDevice) {
```
CarolineLaw, line 47: I'm not sure, but this check might also be in the `MainActivity` and the `SettingsActivity`

PR #265, opened 2020-08-18
Refactor to Navigation component

### LINKED ISSUE #263: Refactor to Navigation component
**Acceptance Criteria**:

All fragment transactions are replaced with Jetpack Navigation components.  No expected change in functionality.

**Design**:

No changes.

jollygreenegiant:
Description: 
Refactor entire app to use the Jetpack Navigation Component.  Part of this involved removing the `SplashActivity` and replacing it with `SplashFragment`, whose activity is `MainActivity`.  The only fragment transactions that remain are for the `FaceTrackFragment`, because I didn't want to make a third navigation graph just for one fragment.

Testing:
None of the UI should have changed, but this touches every screen and every screen change in the app, so we should probably do regression testing.

Fixes #263


app/src/main/java/com/willowtree/vocable/MainActivity.kt
```diff
@@ -31,14 +31,12 @@ class MainActivity : BaseActivity() {
...
-        if (supportFragmentManager.findFragmentById(R.id.fragment_container) == null) {
-            supportFragmentManager
-                .beginTransaction()
-                .replace(R.id.fragment_container, PresetsFragment())
-                .commit()
-        }
+        val navHost = supportFragmentManager.findFragmentById(R.id.main_nav_host_fragment) as NavHostFragment
+        val navController = navHost.navController
+
+        navController.setGraph(R.navigation.main_nav_graph, intent.extras)
```
nicholas-cook, line 37: I don't think you need to call `setGraph` here since you're already setting the `navGraph` property in the activity's xml.

app/src/main/java/com/willowtree/vocable/settings/EditCategoryOptionsFragment.kt
```diff
@@ -37,29 +36,27 @@ class EditCategoryOptionsFragment : BaseFragment<FragmentEditCategoryOptionsBind
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
-        val category = arguments?.getParcelable<Category>(KEY_CATEGORY)
+        // val category = arguments?.get(KEY_CATEGORY) as Category?
+        val category by lazy {
+            fromBundle(requireArguments()).category
```
nicholas-cook, line 41: You can get around having to use `fromBundle` by using the `navArgs` extension.
``` kotlin
private val args: EditCategoryOptionsFragmentArgs by navArgs()
args.category
```

jollygreenegiant: Good catch!  Updated in both places.

PR #268, opened 2020-08-20
Move "hide" button to category list screen

jollygreenegiant:
Description: 
Moves the show/hide button out of the category detail screen and onto the list screen.  It also sets the select buttons on that list so that they're only enabled if the category is user-generated.

- [ ] Regression Testing


app/src/main/res/layout/category_edit_button.xml
```diff
@@ -53,11 +53,23 @@
...
         android:background="@drawable/button_default_background"
         android:src="@drawable/arrow_down_40dp"
+        android:layout_marginEnd="8dp"
         app:layout_constraintBottom_toBottomOf="@id/move_category_up_button"
-        app:layout_constraintEnd_toStartOf="@id/hidden_view"
+        app:layout_constraintEnd_toStartOf="@id/show_hide_category_button"
         app:layout_constraintStart_toEndOf="@id/move_category_up_button"
         app:layout_constraintTop_toTopOf="@id/move_category_up_button" />
 
+    <com.willowtree.vocable.customviews.VocableImageButton
```
nicholas-cook, line 62: We've got three versions of `category_edit_button` so make sure they're all updated if they need to be.

app/src/main/res/layout/fragment_edit_category_options.xml
```diff
@@ -44,52 +44,24 @@
         android:src="@drawable/edit_40dp"
         android:visibility="invisible"
         app:layout_constraintBottom_toBottomOf="@id/category_title"
-        app:layout_constraintEnd_toEndOf="@id/show_category_switch"
+        app:layout_constraintEnd_toEndOf="@id/remove_category_button"
         app:layout_constraintTop_toTopOf="@id/category_title" />
 
-    <com.willowtree.vocable.customviews.VocableConstraintLayout
```
nicholas-cook, line 50: Likewise, there are three versions of this fragment layout.

app/src/main/java/com/willowtree/vocable/settings/EditCategoriesListFragment.kt
```diff
@@ -165,10 +165,28 @@ class EditCategoriesListFragment : BaseFragment<FragmentEditCategoriesListBindin
...
+                editCategorySelectButton.action = {
+                    val action = EditCategoriesFragmentDirections.actionEditCategoriesFragmentToEditCategoryOptionsFragment(category)
+                    findNavController().navigate(action)
+                    editCategoriesViewModel.onCategorySelected(category)
+                }
+            }
+        }
+
+        editButtonBinding.showHideCategoryButton?.apply {
+            if (!category.hidden) {
```
nicholas-cook, line 180: We've been using drawable selectors to handle changing the drawable based on button states. So you can set the image in xml and then the drawable selector will handle toggling the image based on whether we disabled the button above. Take a look at things like `presets_action_button_icon` to see how we've been handling that.

PR #273, opened 2020-08-25
Feature/add phrase to custom category

jollygreenegiant:
Description: 
Adds the ability to add a phrase to a custom category.  The phrase will show up on the main screen, but not under the edit category screen, that will be a separate PR.

- [ ] Regression Testing

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/settings/AddPhraseKeyboardFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/settings/AddPhraseKeyboardFragment.kt
@@ -1,0 +1,83 @@
+package com.willowtree.vocable.settings
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.TextView
+import androidx.core.view.isVisible
+import androidx.lifecycle.Observer
+import androidx.lifecycle.ViewModelProviders
+import androidx.navigation.fragment.findNavController
+import androidx.navigation.fragment.navArgs
+import com.willowtree.vocable.BaseViewModelFactory
+import com.willowtree.vocable.R
+import com.willowtree.vocable.room.Category
+import com.willowtree.vocable.room.Phrase
+import com.willowtree.vocable.settings.EditPhrasesKeyboardFragmentArgs.Companion.fromBundle
+import java.util.*
+
+class AddPhraseKeyboardFragment : EditKeyboardFragment() {
+
+    companion object {
+        private const val KEY_PHRASE = "KEY_PHRASE"
+
+        fun newInstance(phrase: Phrase?): EditPhrasesKeyboardFragment {
+            return EditPhrasesKeyboardFragment().apply {
+                arguments = Bundle().apply {
+                    putParcelable(KEY_PHRASE, phrase)
+                }
+            }
+        }
+    }
+
+
+    private lateinit var viewModel: AddPhraseViewModel
+    private val args: AddPhraseKeyboardFragmentArgs by navArgs()
+    private lateinit var category: Category
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        category = args.category
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        binding.backButton.action = {
+            if (isDefaultTextVisible()) {
+                findNavController().popBackStack()
+            } else {
+                showConfirmationDialog()
+            }
+        }
+
+        binding.saveButton.action = {
+            if (!isDefaultTextVisible()) {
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/willowtree/vocable/settings/AddPhraseViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+class AddPhraseViewModel : BaseViewModel() {
+
+    companion object {
+        private const val PHRASE_UPDATED_DELAY = 2000L
+    }
+
+    private val presetsRepository: PresetsRepository by inject()
+
+    private val liveMySayingsList = MutableLiveData<List<Phrase>>()
+    val mySayingsList: LiveData<List<Phrase>> = liveMySayingsList
```
nicholas-cook, line 24: What are these values going to be used for?

app/src/main/java/com/willowtree/vocable/settings/EditCategoryOptionsFragment.kt
```diff
@@ -46,7 +46,7 @@ class EditCategoryOptionsFragment : BaseFragment<FragmentEditCategoryOptionsBind
             findNavController().popBackStack()
         }
 
-        binding.removeCategoryButton.action = {
+        binding.removeCategoryButton.setOnClickListener {
```
nicholas-cook, line 49: We should use `action` here instead of setting a click listener. The `action` is what will get triggered by head tracking and it will get automatically triggered by a user clicking the screen.

app/src/main/java/com/willowtree/vocable/settings/EditCategoryOptionsFragment.kt
```diff
@@ -67,6 +67,11 @@ class EditCategoryOptionsFragment : BaseFragment<FragmentEditCategoryOptionsBind
             }
         }
 
+        binding.addPhraseButton.setOnClickListener {
```
nicholas-cook, line 70: Same comment on `action`

app/src/main/java/com/willowtree/vocable/settings/AddPhraseKeyboardFragment.kt
```diff
@@ -0,0 +1,84 @@
...
+            } else {
+                showConfirmationDialog()
+            }
+        }
+
+        binding.saveButton.action = {
+            if (!isDefaultTextVisible()) {
+                binding.keyboardInput.text.let { text ->
+                    if (text.isNotBlank()) {
+                        viewModel.addNewPhrase(text.toString(), category.categoryId)
```
nicholas-cook, line 60: Are we going to add the ability to update the phrase here? It doesn't need to be done in this PR, but we need to make sure we don't forget about it.

app/src/main/res/drawable/ic_plus_40dp.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:height="40dp" android:tint="#D1EEE9"
```
nicholas-cook, line 1: Is this icon being used?

app/src/main/java/com/willowtree/vocable/settings/AddPhraseKeyboardFragment.kt
```diff
@@ -0,0 +1,84 @@
...
+                }
+            }
+        }
+
+        (binding.phraseSavedView.root as TextView).apply {
+            setText(R.string.new_phrase_saved)
+        }
+
+        viewModel = ViewModelProviders.of(
+            requireActivity(),
```
nicholas-cook, line 72: If this view model is only going to be used by this fragment, you can pass in `this` instead of `requireActivity`

app/src/main/java/com/willowtree/vocable/settings/AddPhraseViewModel.kt
jollygreenegiant: These were holdovers from a copy/paste, they need to be removed I think

app/src/main/java/com/willowtree/vocable/settings/EditCategoryOptionsFragment.kt
jollygreenegiant: Gotcha - I switched this view from a ConstraintLayout to a VocableImageButton and it didn't like the action keyword.  Could have just been an IDE issue, I'll look into it

app/src/main/res/drawable/ic_plus_40dp.xml
jollygreenegiant: This was part of my original UI changes, I'll remove it

app/src/main/java/com/willowtree/vocable/settings/AddPhraseKeyboardFragment.kt
jollygreenegiant: I think it should be separate, I'll make an issue

Issue #239, opened 2020-05-21
Add/Edit/Delete Custom Categories


hlargen5:
**As a user, I would like to create new custom categories, edit their names later, and delete entire categories to enhance the customization of my experience.**

**Acceptance Criteria:**

- To add a new category, the user will tap the "plus" icon in the top right-hand corner of the Categories screen. This will take them into the add keyboard flow. New categories will be added to the bottom of the categories list, and should appear on the home screen.
- If a user closes the keyboard without saving, they should receive the discard edits alert.
- To edit a custom category name, the user will navigate to the category detail screen and tap the pencil icon. This will take them into the edit keyboard flow.
- If a user closes the keyboard without saving any changes, they should receive the discard edits alert.
- If a user does not make any changes and closes the keyboard, no alert should be given.
- To delete an entire category, a user should navigate to the custom category detail screen and tap the trash can icon in the top right-hand corner. This will delete the entire category and all of its custom phrases in both the Categories setting screen and the home screen.
- A user should receive the "are you sure you want to delete?" alert when tapping the delete category button.

**Test Note:**
Custom Categories will be in English only for this ticket.

**Design:**

-https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable-New?node-id=19%3A1

Issue #182, opened 2020-04-27
Edit Phrase for custom category


Uma-mv:
**As a user, I need the ability to edit existing custom phrases to maximize the customization of my experience.**

**AC**

- User will have the ability to view existing phrases within the custom category and enter an edit flow to update the phrase
- Leaving the flow without changes should allow the user to exit without a warning
- Attempting to exit without saving changes should prompt the user to save those changes.

**Design:**
https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable-New?node-id=19%3A1

PR #275, opened 2020-08-26
Display a custom category's phrases

nicholas-cook:
Description: 
- Adds logic to display a custom category's phrases
- Adds a `RecyclerView` adapter to handle inflating and displaying phrases
- Adds layouts for the adapter's items
- Creates pager adapter for handling paging between a category's phrases

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/settings/EditCategoriesViewModel.kt
+++ b/app/src/main/java/com/willowtree/vocable/settings/EditCategoriesViewModel.kt
@@ -5,6 +5,7 @@ import androidx.lifecycle.MutableLiveData
 import com.willowtree.vocable.BaseViewModel
 import com.willowtree.vocable.presets.PresetsRepository
 import com.willowtree.vocable.room.Category
+import com.willowtree.vocable.room.Phrase
 import com.willowtree.vocable.utils.LocalizedResourceUtility
 import kotlinx.coroutines.launch
 import org.koin.core.inject
@@ -21,6 +22,9 @@ class EditCategoriesViewModel : BaseViewModel() {
 
     private val liveLastViewedIndex = MutableLiveData<Int>()
     val lastViewedIndex: LiveData<Int> = liveLastViewedIndex
+
+    private val liveCategoryPhraseList = MutableLiveData<List<Phrase>>()
+    val categoryPhraseList: LiveData<List<Phrase>> = liveCategoryPhraseList
 
     private var overallCategories = listOf<Category>()
 
@@ -56,6 +60,7 @@ class EditCategoriesViewModel : BaseViewModel() {
     fun deleteCategory(category: Category) {
         backgroundScope.launch {
             presetsRepository.deleteCategory(category)
+            // TODO: Delete category's phrases too
             refreshCategories()
         }
     }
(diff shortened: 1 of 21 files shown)


app/src/main/java/com/willowtree/vocable/settings/EditCategoriesViewModel.kt
```diff
@@ -56,6 +60,7 @@ class EditCategoriesViewModel : BaseViewModel() {
     fun deleteCategory(category: Category) {
         backgroundScope.launch {
             presetsRepository.deleteCategory(category)
+            // TODO: Delete category's phrases too
```
jollygreenegiant, line 63: We should make an issue for this if there isn't already

app/src/main/java/com/willowtree/vocable/settings/customcategories/CustomCategoryPhraseListFragment.kt
```diff
@@ -0,0 +1,63 @@
...
+
+        fun newInstance(phrases: List<Phrase>): CustomCategoryPhraseListFragment {
+            return CustomCategoryPhraseListFragment().apply {
+                arguments = bundleOf(KEY_PHRASES to ArrayList(phrases))
+            }
+        }
+    }
+
+    private val onPhraseEdit = { phrase: Phrase ->
+        // TODO: Handle editing the phrase
```
jollygreenegiant, line 28: We should make an issue for this too

app/src/main/java/com/willowtree/vocable/settings/customcategories/adapter/CustomCategoryPhraseAdapter.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    inner class CustomCategoryPhraseViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
+        private val binding: EditCustomCategoryPhraseItemBinding =
+            EditCustomCategoryPhraseItemBinding.bind(itemView)
+
+        fun bind(phrase: Phrase, onPhraseEdit: (Phrase) -> Unit, onPhraseDelete: (Phrase) -> Unit) {
+            binding.phraseText.text = localizedResourceUtility.getTextFromPhrase(phrase)
+
+            with(binding.actionButtonContainer) {
+                editSayingsButton.action = {
```
jollygreenegiant, line 35: We might want to rename these buttons to use "phrase" instead of "sayings", for consistency

app/src/main/java/com/willowtree/vocable/settings/customcategories/CustomCategoryPhraseListFragment.kt
nicholas-cook: Looks like that's issue #182

app/src/main/java/com/willowtree/vocable/settings/EditCategoriesViewModel.kt
nicholas-cook: Hmm so deleting the category is part of #239, should we split it out?

nicholas-cook: Opened #276

PR #280, opened 2020-08-27
Custom category empty state

### LINKED ISSUE #278: Custom categories empty state
**Acceptance Criteria**:

When a user views a custom category that has no phrases in it, they should be shown a message, and given a button to add phrases to that category.

jollygreenegiant:
Description: 
Shows the empty state for a custom category when the list of phrases comes back empty.  Allows the user to add a new phrase from the empty state.

Closes #278 

- [ ] Regression Testing


app/src/main/java/com/willowtree/vocable/settings/EditCategoryOptionsFragment.kt
```diff
@@ -176,23 +181,30 @@ class EditCategoryOptionsFragment : BaseFragment<FragmentEditCategoryOptionsBind
...
-                setCurrentItem(
-                    middle + (phrasesAdapter.numPages - mod),
-                    false
-                )
+        if (phrases.isEmpty()) {
+            binding.emptyPhrasesText.visibility = View.VISIBLE
+            binding.emptyAddPhraseButton.visibility = View.VISIBLE
+        } else {
+            binding.emptyPhrasesText.visibility = View.GONE
+            binding.emptyAddPhraseButton.visibility = View.GONE
```
nicholas-cook, line 189: There's actually a Kotlin Android extension to help with toggling between `VISIBLE` and `GONE`:
`binding.emptyPhrasesText.isVisible = true`

app/src/main/res/layout-land/fragment_edit_category_options.xml
```diff
@@ -70,6 +70,42 @@
...
         app:layout_constraintTop_toTopOf="@+id/edit_options_back_button" />
 
+    <TextView
+        android:id="@+id/empty_phrases_text"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:text="@string/custom_category_empty"
+        android:textColor="@color/textColor"
+        android:textStyle="bold"
+        android:textSize="22sp"
```
nicholas-cook, line 80: We're trying to avoid hardcoded values in our layout files. Could you move these to the corresponding `dimens` files? There should be one for each form factor. It might also be worth making a style if these all share similar properties.

app/src/main/java/com/willowtree/vocable/settings/EditCategoryOptionsFragment.kt
jollygreenegiant: 🙏

app/src/main/res/values/dimens.xml
```diff
@@ -262,5 +262,7 @@
     <!-- Category Edit Button -->
     <dimen name="category_edit_button_hidden_margin_start">16dp</dimen>
     <dimen name="category_edit_button_hidden_margin_end">8dp</dimen>
+    <dimen name="custom_category_button_text_size_phone">22sp</dimen>
+    <dimen name="custom_category_button_text_size_tablet">34sp</dimen>
```
nicholas-cook, line 266: So we actually have a separate `dimens` file for tablets (and also landscape). It's set up the same way as the layout files, in directories named with `sw600dp`, `land`, etc. That way you only have to have one style that will pull dimens from the correct resource set.

jollygreenegiant: I knew there had to be a way to do that, I was looking for different styles files.  I really need to work with different screen formats more

jennydoesthings:
The message is showing up as expected, but the "Add Phrase" button is not accepting clicks.

PR #282, opened 2020-08-28
Add empty state to main page custom category

### LINKED ISSUE #281: Custom Category empty state 2.0
**Acceptance Criteria**:

On the main screen, an empty custom category should show info text with an Add Phrase button, same as on the detail screen.

jollygreenegiant:
Description: 
Adds an empty state for custom category to the main presets screen.  Right now, the Add Phrase button will bring you to the keyboard, but saving the phrase from there will add the phrase to My Sayings, not the category you came from.  This will be changed in a future PR, where we will add the page with multiple switches to choose which categories a new phrase goes into.

Closes #281 

- [ ] Regression Testing

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
@@ -24,6 +24,7 @@ import com.willowtree.vocable.settings.SettingsActivity
 import com.willowtree.vocable.utils.SpokenText
 import com.willowtree.vocable.utils.VocableFragmentStateAdapter
 import com.willowtree.vocable.utils.VocableTextToSpeech
+import kotlinx.android.synthetic.main.fragment_presets.*
 
 class PresetsFragment : BaseFragment<FragmentPresetsBinding>() {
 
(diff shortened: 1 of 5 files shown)


app/src/main/res/layout-land/fragment_presets.xml
```diff
@@ -85,6 +85,32 @@
...
+        app:layout_constraintWidth_percent=".75"
+        app:layout_constraintTop_toTopOf="@id/phrases_view"
+        app:layout_constraintBottom_toTopOf="@id/empty_add_phrase_button"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"/>
+
+    <com.willowtree.vocable.customviews.KeyboardButton
+        android:id="@+id/empty_add_phrase_button"
+        android:layout_width="0dp"
+        android:layout_height="64dp"
```
nicholas-cook, line 104: Don't forget to move these heights to a corresponding `dimens` file. It looks like they're all `64dp`, so you can just add it to the base `dimens` file.

app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
```diff
@@ -93,6 +94,11 @@ class PresetsFragment : BaseFragment<FragmentPresetsBinding>() {
             findNavController().navigate(R.id.action_presetsFragment_to_settingsActivity)
         }
 
+        binding.emptyAddPhraseButton.action = {
+            val action = PresetsFragmentDirections.actionPresetsFragmentToKeyboardFragment()
```
nicholas-cook, line 98: We'll probably want to pass the current category to the keyboard fragment so we can set that category to checked on [this screen](https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable-New?node-id=136%3A466). We can do that in another PR though since that flow will come later.

jollygreenegiant: Yeah my plan is to do this with the PR that adds the screen with the switches.

PR #284, opened 2020-08-28
Delete phrase from custom category

nicholas-cook:
Description: 
- Adds logic to allow user to delete a phrase from a custom category directly from the list of phrases
- The phrase will be totally deleted from the database if it is only associated with the category currently being viewed. Otherwise it will remain in the database for its other categories
- [Figma link](https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable-New?node-id=122%3A2791)

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/settings/customcategories/CustomCategoryPhraseListFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/settings/customcategories/CustomCategoryPhraseListFragment.kt
@@ -3,12 +3,17 @@ package com.willowtree.vocable.settings.customcategories
 import android.os.Bundle
 import android.view.View
 import androidx.core.os.bundleOf
+import androidx.core.view.isVisible
+import androidx.lifecycle.ViewModelProviders
 import androidx.recyclerview.widget.GridLayoutManager
 import com.willowtree.vocable.BaseFragment
+import com.willowtree.vocable.BaseViewModelFactory
 import com.willowtree.vocable.BindingInflater
 import com.willowtree.vocable.R
 import com.willowtree.vocable.databinding.FragmentCustomCategoryPhraseListBinding
+import com.willowtree.vocable.room.Category
 import com.willowtree.vocable.room.Phrase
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/willowtree/vocable/settings/customcategories/CustomCategoryPhraseListFragment.kt
```diff
@@ -3,33 +3,45 @@ package com.willowtree.vocable.settings.customcategories
...
+        private const val KEY_CATEGORY = "KEY_CATEGORY"
 
-        fun newInstance(phrases: List<Phrase>): CustomCategoryPhraseListFragment {
+        fun newInstance(
+            phrases: List<Phrase>,
+            category: Category
+        ): CustomCategoryPhraseListFragment {
             return CustomCategoryPhraseListFragment().apply {
-                arguments = bundleOf(KEY_PHRASES to ArrayList(phrases))
+                arguments = bundleOf(KEY_PHRASES to ArrayList(phrases), KEY_CATEGORY to category)
```
jollygreenegiant, line 31: Do we still need this, now that we're using Navigation?  New fragment arguments should be added to that.

nicholas-cook: So this is a fragment that we're adding to a `FragmentStateAdapter`, so we're not technically navigating to it, we're just using it to populate an adapter. I don't know if we can use the nav component for that.

jennydoesthings:
When I delete the last phrase in a category, the page numbers at the bottom go a little weird.


Screen Shot 2020-08-31 at 3 10 44 PM

jennydoesthings:
> When I delete the last phrase in a category, the page numbers at the bottom go a little weird.

Actually looks like Cameron resolved this in #285

SashankPatel:
Verified phrases can be deleted individually.
Verified Confirmation modal popup.
Verified, if the same phrase is in multiple categories, they can be deleted independently, and is reflected accurately on the Home screen.

PR #287, opened 2020-09-02
Add custom phrase from keyboard

### LINKED ISSUE #241: Add Custom Phrases to Custom Categories from Keyboard
**As a user, if I type out a new phrase on the keyboard, I would like to be able to easily add it to my custom categories without going to the Settings screen.**

**Acceptance Criteria:**

- The star icon (previously added to "My Sayings") should be changed to a "plus" icon, which opens a "Add to Category" screen.
- On the "Add to Category" screen, users should see all of their custom categories that they created, and can toggle on and off which categories the phrase should be added to.
- If a user adds a phrase to a category, they should receive a "Saved to [Category Name]" Alert modal
- If a user removes a phrase from a category, they should receive a "Removed from [Category Name]" Alert Modal.
- If there are no custom categories, display the empty state instead.

**Design:**
https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable-New?node-id=35%3A5

jollygreenegiant:
We need to double check that the new/updated strings are updated by Crowdin when this is merged.

Description: 
Adds the ability to save a phrase to a custom category from the main keyboard, rather than automatically adding it to My Sayings.  This essentially renders My Sayings obsolete, and that will be removed as part of a separate issue.  This also does not cover the empty state for this, which is a separate issue as well.

Closes #241 

- [ ] Regression Testing

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
@@ -91,9 +91,7 @@ class AddToCategoryPickerFragment : BaseFragment<FragmentAddToCategoryPickerBind
 
         subscribeToViewModel()
 
-        with(addToCategoryPickerViewModel) {
-            getCategoryList()
-        }
+        addToCategoryPickerViewModel.getCategoryList()
     }
 
     private fun subscribeToViewModel() {
--- a/app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
@@ -29,11 +29,12 @@ class AddToCategoryPickerFragment : BaseFragment<FragmentAddToCategoryPickerBind
 
     private var maxCategories = 1
     private lateinit var categoriesAdapter: CategoriesPagerAdapter
+    private lateinit var phraseText: String
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
-        val phraseText = args.phraseText
+        phraseText = args.phraseText
 
         binding.backButton.action = {
             findNavController().popBackStack()
@@ -91,45 +92,13 @@ class AddToCategoryPickerFragment : BaseFragment<FragmentAddToCategoryPickerBind
         subscribeToViewModel()
 
         with(addToCategoryPickerViewModel) {
-            buildCategoryList(phraseText)
+            getCategoryList()
         }
     }
 
     private fun subscribeToViewModel() {
-        addToCategoryPickerViewModel.categoryMap.observe(viewLifecycleOwner, Observer {
-            it?.let { map ->
-                with(ArrayList(map.keys)) {
-                    val categoriesExist = this.size != 0
-
-                    binding.emptyStateTitle.isVisible = !categoriesExist
-                    binding.emptyStateText.isVisible = !categoriesExist
-                    binding.categoryPagerBackButton.isEnabled = categoriesExist
-                    binding.categoryPagerForwardButton.isEnabled = categoriesExist
-
-                    // empty state
-                    if (this.size == 0) {
-                        binding.categoryPageNumber.text = getString(R.string.phrases_page_number, 1, 1)
-                    } else {
-                        binding.categoryHolder.apply {
-                            isSaveEnabled = false
-                            adapter = categoriesAdapter
-                            categoriesAdapter.setItems(this@with)
-
-                            // Move adapter to middle so user can scroll both directions
-                            val middle = categoriesAdapter.itemCount / 2
-                            if (middle % categoriesAdapter.numPages == 0) {
-                                setCurrentItem(middle, false)
-                            } else {
-                                val mod = middle % categoriesAdapter.numPages
-                                setCurrentItem(
-                                    middle + (categoriesAdapter.numPages - mod),
-                                    false
-                                )
-                            }
-                        }
-                    }
-                }
-            }
+        addToCategoryPickerViewModel.categoryList.observe(viewLifecycleOwner, Observer {
+            handleCategories(it)
         })
     }
 
@@ -137,8 +106,40 @@ class AddToCategoryPickerFragment : BaseFragment<FragmentAddToCategoryPickerBind
         return emptyList()
     }
 
+    private fun handleCategories(categories: List<Category>) {
+        val categoriesExist = categories.isNotEmpty()
+
+        binding.emptyStateTitle.isVisible = !categoriesExist
+        binding.emptyStateText.isVisible = !categoriesExist
+        binding.categoryPagerBackButton.isEnabled = categoriesExist
+        binding.categoryPagerForwardButton.isEnabled = categoriesExist
+
+        if (categoriesExist) {
+            with(binding.categoryHolder) {
+                isSaveEnabled = false
+
+                adapter = categoriesAdapter
+
+                categoriesAdapter.setItems(categories)
+
+                // Move adapter to middle so user can scroll both directions
+                val middle = categoriesAdapter.itemCount / 2
+                if (middle % categoriesAdapter.numPages == 0) {
+                    setCurrentItem(middle, false)
+                } else {
+                    val mod = middle % categoriesAdapter.numPages
+                    setCurrentItem(
+                        middle + (categoriesAdapter.numPages - mod),
+                        false
+                    )
+                }
+            }
+        }
+    }
+
     inner class CategoriesPagerAdapter(fm: FragmentManager) :
             VocableFragmentStateAdapter<Category>(fm, viewLifecycleOwner.lifecycle) {
+
         override fun setItems(items: List<Category>) {
             super.setItems(items)
             setPagingButtonsEnabled(numPages > 1)
--- a/app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
@@ -102,12 +102,15 @@ class AddToCategoryPickerFragment : BaseFragment<FragmentAddToCategoryPickerBind
         addToCategoryPickerViewModel.categoryMap.observe(viewLifecycleOwner, Observer {
             it?.let { map ->
                 with(ArrayList(map.keys)) {
+                    val categoriesExist = this.size != 0
+
+                    binding.emptyStateTitle.isVisible = !categoriesExist
+                    binding.emptyStateText.isVisible = !categoriesExist
+                    binding.categoryPagerBackButton.isEnabled = categoriesExist
+                    binding.categoryPagerForwardButton.isEnabled = categoriesExist
+
                     // empty state
                     if (this.size == 0) {
-                        binding.emptyStateTitle.isVisible = true
-                        binding.emptyStateText.isVisible = true
-                        binding.categoryPagerBackButton.isEnabled = false
-                        binding.categoryPagerForwardButton.isEnabled = false
                         binding.categoryPageNumber.text = getString(R.string.phrases_page_number, 1, 1)
                     } else {
                         binding.categoryHolder.apply {
--- a/app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
@@ -1,0 +1,153 @@
+package com.willowtree.vocable.presets
+
+import android.os.Bundle
+import android.view.View
+import android.widget.TextView
+import androidx.core.view.isVisible
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentManager
+import androidx.lifecycle.Observer
+import androidx.lifecycle.ViewModelProviders
+import androidx.navigation.fragment.findNavController
+import androidx.navigation.fragment.navArgs
+import androidx.viewpager2.widget.ViewPager2
+import com.willowtree.vocable.BaseFragment
+import com.willowtree.vocable.BaseViewModelFactory
+import com.willowtree.vocable.BindingInflater
+import com.willowtree.vocable.R
+import com.willowtree.vocable.databinding.FragmentAddToCategoryPickerBinding
+import com.willowtree.vocable.presets.adapter.AddToCategoryPickerViewModel
+import com.willowtree.vocable.room.Category
+import com.willowtree.vocable.settings.SettingsActivity
+import com.willowtree.vocable.utils.VocableFragmentStateAdapter
+import kotlin.math.min
+
+class AddToCategoryPickerFragment : BaseFragment<FragmentAddToCategoryPickerBinding>() {
+
+    private val args: AddToCategoryPickerFragmentArgs by navArgs()
+
+    override val bindingInflater: BindingInflater<FragmentAddToCategoryPickerBinding> =
+        FragmentAddToCategoryPickerBinding::inflate
+    private lateinit var addToCategoryPickerViewModel: AddToCategoryPickerViewModel
+
+    private var maxCategories = 1
+    private lateinit var categoriesAdapter: CategoriesPagerAdapter
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        val phraseText = args.phraseText
+
+        binding.backButton.action = {
+            findNavController().popBackStack()
+        }
+
+        val numColumns = resources.getInteger(R.integer.custom_category_columns)
+        val numRows = resources.getInteger(R.integer.custom_category_rows)
+        maxCategories = numColumns * numRows
+
+        categoriesAdapter = CategoriesPagerAdapter(childFragmentManager)
+
+        binding.categoryPagerForwardButton.action = {
+            when (val currentPosition = binding.categoryHolder.currentItem) {
+                categoriesAdapter.itemCount - 1 -> {
+                    binding.categoryHolder.setCurrentItem(0, true)
+                }
+                else -> {
+                    binding.categoryHolder.setCurrentItem(currentPosition + 1, true)
+                }
+            }
+        }
+
+        binding.categoryPagerBackButton.action = {
+            when (val currentPosition = binding.categoryHolder.currentItem) {
+                0 -> {
+                    binding.categoryHolder.setCurrentItem(
+                        categoriesAdapter.itemCount - 1,
+                        true
+                    )
+                }
+                else -> {
+                    binding.categoryHolder.setCurrentItem(currentPosition - 1, true)
+                }
+            }
+        }
+
+        binding.categoryHolder.registerOnPageChangeCallback(object :
+            ViewPager2.OnPageChangeCallback() {
+            override fun onPageSelected(position: Int) {
+                val pageNum = position % categoriesAdapter.numPages + 1
+                binding.categoryPageNumber.text = getString(R.string.phrases_page_number, pageNum, categoriesAdapter.numPages)
+
+                val activity = requireActivity()
+                if (activity is SettingsActivity) {
+                    activity.resetAllViews()
+                }
+            }
+        })
+
+        addToCategoryPickerViewModel = ViewModelProviders.of(
+            requireActivity(),
+            BaseViewModelFactory()
+        ).get(AddToCategoryPickerViewModel::class.java)
+
+        subscribeToViewModel()
+
+        with(addToCategoryPickerViewModel) {
+            buildCategoryList(phraseText)
+        }
+    }
+
+    private fun subscribeToViewModel() {
+        addToCategoryPickerViewModel.categoryMap.observe(viewLifecycleOwner, Observer {
+            it?.let { map ->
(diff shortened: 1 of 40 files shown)


jollygreenegiant:
I'm treating this as something of a first pass.  It has all the functionality, but I think there may be places where the logic is complex and may be able to be improved upon after a review.

app/src/main/java/com/willowtree/vocable/presets/adapter/AddToCategoryPickerViewModel.kt
```diff
@@ -0,0 +1,158 @@
...
+
+            // for each category in the list
+            for (category in userCategories) {
+                // get all the phrases associated with it
+                val crossRefs = presetsRepository.getCrossRefsForCategoryId(category.categoryId)
+
+                // for each phrase in the category
+                for (ref in crossRefs) {
+                    // get the phrase from the db
+                    val phrase = presetsRepository.getPhraseById(ref.phraseId)
```
nicholas-cook, line 65: There's already a call to get all the phrases for a given category - `presetsRepository.getPhrasesForCategory(categoryId: String)`

app/src/main/java/com/willowtree/vocable/presets/adapter/AddToCategoryPickerViewModel.kt
```diff
@@ -0,0 +1,158 @@
...
+
+    fun buildCategoryList(phraseString: String) {
+        // map category to whether or not the phrase is in that category
+        val phraseMap = mutableMapOf<Category, Boolean>()
+
+        backgroundScope.launch {
+            userCategories = presetsRepository.getUserGeneratedCategories()
+
+            // for each category in the list
+            for (category in userCategories) {
```
nicholas-cook, line 58: You can use Kotlin's `forEach` extension here and other places you have for-loops - `userCategories.forEach {}`

SashankPatel:
Bug 1: Sept 3: 
I did a fresh instsall of the app and ran it and found this: 
Screen Shot 2020-09-03 at 7 46 50 AM

SashankPatel:
Sept 3rd: Bug 2: 

1. Launch the app
2. Tap on the Keyboard icon.
3. Tap on the Delete button

Expected: Nothing to happen
Actual: App crashes.

E/AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-2
    Process: com.willowtree.vocable, PID: 17009
    java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.List com.willowtree.vocable.room.CategoryWithPhrases.getPhrases()' on a null object reference
        at com.willowtree.vocable.presets.PresetsRepository.getPhrasesForCategory(PresetsRepository.kt:24)
        at com.willowtree.vocable.presets.PresetsRepository$getPhrasesForCategory$1.invokeSuspend(Unknown Source:12)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(Dispatched.kt:241)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:594)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.access$runSafely(CoroutineScheduler.kt:60)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:740)
I/Process: Sending signal. PID: 17009 SIG: 9

SashankPatel:
Sept 3rd Bug 1: 

Device: Pixel 2

Scenario:
1.) Hold the phone in Portrait View.
2.) Launch the app
3.) Add a New Category without phrases
4.) Navigate to the Home screen and select newly created category.
5.) Rotate the app

Expected: 
Expected_EmptyState

Actual:
Actual_EmptyState

SashankPatel:
Question: PR Description notes Empty state is not covered in this PR, it is a separate issue.  The AC on 241 mentions Empty state to be displayed, does the AC need updating?  Is there.a ticket number for implementing empty state?

SashankPatel:
Sept 3rd Bug 2: App Crash:

Device: Pixel 2

Scenario:
1.) Launch the app
2.) Go to Settings and Create a new Custom Category
3.) Go back to the Home Screen and Select the newly created category
4.) Click on the Add Phrase button.
5.) Click on the Settings button.
6.) Navigate to the newly created custom Category
7.) Delete the category
8.) Tap on the back button.
9.) Tap on the "X" to close the Settings View.
10.) Tap on the "Comment bubble" icon to go back to the home screen

Expected: Home screen to be displayed.
Acutal: App Crashes.

E/AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-1
    Process: com.willowtree.vocable, PID: 19335
    java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.List com.willowtree.vocable.room.CategoryWithPhrases.getPhrases()' on a null object reference
        at com.willowtree.vocable.presets.PresetsRepository.getPhrasesForCategory(PresetsRepository.kt:24)
        at com.willowtree.vocable.presets.PresetsViewModel$onCategorySelected$1.invokeSuspend(PresetsViewModel.kt:40)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(Dispatched.kt:241)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:594)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.access$runSafely(CoroutineScheduler.kt:60)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:740)
I/Process: Sending signal. PID: 19335 SIG: 9

app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        binding.categoryHolder.registerOnPageChangeCallback(object :
+            ViewPager2.OnPageChangeCallback() {
+            override fun onPageSelected(position: Int) {
+                val pageNum = position % categoriesAdapter.numPages + 1
+                binding.categoryPageNumber.text = getString(R.string.phrases_page_number, pageNum, categoriesAdapter.numPages)
+
+                val activity = requireActivity()
+                if (activity is SettingsActivity) {
```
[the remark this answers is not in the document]
CarolineLaw: is there a way to get to this fragment from the `MainActivity`?

app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerListFragment.kt
```diff
@@ -0,0 +1,112 @@
...
+                )
+            }
+        }
+    }
+
+    private val args: AddToCategoryPickerFragmentArgs by navArgs()
+
+    private lateinit var viewModel: AddToCategoryPickerViewModel
+    private lateinit var phraseString: String
+    private lateinit var categories: List<Category>
```
CarolineLaw, line 44: are these categories used anywhere? might be able to get rid of it maybe?

SashankPatel:
> Question: PR Description notes Empty state is not covered in this PR, it is a separate issue. The AC on 241 mentions Empty state to be displayed, does the AC need updating? Is there.a ticket number for implementing empty state?

Resolved: Keyboard - Add Phrase without Custom Category Empty State is implemented in commit: b836f87

jollygreenegiant:
Bug one pulled out into a new issue: #288

app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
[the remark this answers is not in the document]
jollygreenegiant: It is coming from the main keyboard, so this fragment would be under MainActivity.  I'll update it.

app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerListFragment.kt
```diff
@@ -0,0 +1,112 @@
...
+        ).get(AddToCategoryPickerViewModel::class.java)
+
+
+
+        subscribeToViewModel()
+    }
+
+    private fun subscribeToViewModel() {
+        viewModel.showPhraseAdded.observe(viewLifecycleOwner, Observer {
+            binding.phraseSavedView.root.isVisible = it
```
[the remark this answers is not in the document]
jollygreenegiant: It doesn't directly handle it, but it's behaving correctly - the type of this LiveDate is a non-nullable Boolean, so it should default to false, not null

app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
```diff
@@ -0,0 +1,162 @@
...
+        binding.categoryPagerForwardButton.action = {
+            when (val currentPosition = binding.categoryHolder.currentItem) {
+                categoriesAdapter.itemCount - 1 -> {
+                    binding.categoryHolder.setCurrentItem(0, true)
+                }
+                else -> {
+                    binding.categoryHolder.setCurrentItem(currentPosition + 1, true)
+                }
+            }
+        }
```
luonglegs, line 57: Pretty cool to pass in lambdas into `action`! I think the one downside of this would be readability that we lose because it doesn't often promote encapsulation. It would be more readable if we we set a member val to hold the lambda and summarize its function.

```
val onForwardButtonClick = {
            when (val currentPosition = binding.categoryHolder.currentItem) {
                categoriesAdapter.itemCount - 1 -> {
                    binding.categoryHolder.setCurrentItem(0, true)
                }
                else -> {
                    binding.categoryHolder.setCurrentItem(currentPosition + 1, true)
                }
            }
        }
...

// Then in onViewCreated you can set it
binding.categoryPagerForwardButton.action = onForwardButtonClick
```

This can be done for all the other actions as well. Maybe this is a paradigm your project has already been promoting, if that's the case then I think it's fine. Just jarring for me to have to read thru all of that in 1 method.

app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
```diff
@@ -0,0 +1,162 @@
...
+            override fun onPageSelected(position: Int) {
+                val pageNum = position % categoriesAdapter.numPages + 1
+                binding.categoryPageNumber.text = getString(R.string.phrases_page_number, pageNum, categoriesAdapter.numPages)
+
+                val activity = requireActivity()
+                if (activity is MainActivity) {
+                    activity.resetAllViews()
+                }
+            }
+        })
```
luonglegs, line 84: Since you only care about `onPageSelected(position: Int)`, you could probably create an extension function to make this more readable and re-usable elsewhere. Something like...

```
fun ViewPager2.onPageSelected(onPageSelected: (Int) -> Unit) {
    registerOnPageChangeCallback(object : ViewPager2.OnPageChangeCallback() {
        override fun onPageSelected(position: Int) {
            super.onPageSelected(position)
            onPageSelected(position)
        }
    })
}
...

// So then you could just call something like
binding.categorHolder.onPageSelected { position ->
    // Do your thangs here
}
```

Or since yall like using lambdas and what not. You could even do some sort of extension val instead.
```
binding.categoryHolder.onPageSelected = { position ->
    // Do thangs
}
```

app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
```diff
@@ -0,0 +1,162 @@
...
+        addToCategoryPickerViewModel = ViewModelProviders.of(
+            requireActivity(),
+            BaseViewModelFactory()
+        ).get(AddToCategoryPickerViewModel::class.java)
+
+        subscribeToViewModel()
+
+        with(addToCategoryPickerViewModel) {
+            buildCategoryList(phraseText)
+        }
```
luonglegs, line 95: Is there a benefit or point to using the "with" here?

app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
```diff
@@ -0,0 +1,162 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+        })
+    }
+
+    override fun getAllViews(): List<View> {
+        return emptyList()
```
luonglegs, line 137: Is this supposed to return an empty list?

app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
```diff
@@ -0,0 +1,162 @@
...
+                            if (middle % categoriesAdapter.numPages == 0) {
+                                setCurrentItem(middle, false)
+                            } else {
+                                val mod = middle % categoriesAdapter.numPages
+                                setCurrentItem(
+                                    middle + (categoriesAdapter.numPages - mod),
+                                    false
+                                )
+                            }
+                        }
```
luonglegs, line 129: So this is probably another example of where extracting all of this logic out into its own benefits the code base bc it makes it more readable which also lessens the chances for mistakes. It'd also be easier to find em too. There are a couple of KTX extension functions that help out with this too.

So usual on my projects we'd have a method just like your `subscribeToViewModel()` except ours is a lil smaller and intentional. Each LiveData is being observed and firing off their own methods to handle.
```
private fun observeViewModel() = viewModel.apply {
        recommendation.observe(viewLifecycleOwner, ::bindRecommendation)
        feedback.observe(viewLifecycleOwner, ::bindFeedback)
        error.observe(viewLifecycleOwner, ::handleError)
    }

private fun bindRecommendation(recommendation: LolaRecommendation) {
    // Do thangs
}
```

app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
```diff
@@ -0,0 +1,162 @@
...
+                categoryPagerForwardButton.isEnabled = enable
+            }
+        }
+
+        override fun getMaxItemsPerPage(): Int = maxCategories
+
+        override fun createFragment(position: Int): Fragment {
+            val categories = getItemsByPosition(position)
+
+            return AddToCategoryPickerListFragment.newInstance(args.phraseText, categories)
```
luonglegs, line 159: Is `args.phraseText` ever null? If it is, is that handled?

app/src/main/java/com/willowtree/vocable/presets/adapter/AddToCategoryPickerViewModel.kt
```diff
@@ -0,0 +1,154 @@
...
+                        // add it to the map
+                        phraseMap[category] = true
+                    }
+                }
+
+                // if we get through the loop without adding anything, the phrase isn't in there
+                if (phraseMap[category] == null) {
+                    phraseMap[category] = false
+                }
+            }
```
luonglegs, line 75: This is in a `backgroundScope` so i'm not too worried about it but do you see any bad performance issues w/ this being a nested loop? Is there any way around this? I don't have enuff context on the project but i always try to avoid nested loops as much as I can.

app/src/main/java/com/willowtree/vocable/presets/adapter/AddToCategoryPickerViewModel.kt
```diff
@@ -0,0 +1,154 @@
...
+    private fun deletePhraseFromCategory(phraseString: String, category: Category) {
+        backgroundScope.launch {
+            var phrase: Phrase? = null
+
+            val phrasesInCategory = presetsRepository.getPhrasesForCategory(category.categoryId)
+            for (p in phrasesInCategory) {
+                if (localizedResourceUtility.getTextFromPhrase(p) == phraseString) {
+                    phrase = p
+                }
+            }
```
luonglegs, line 90: You could probably just use a `findLast {  }` function to set `phrase` here.

```
val phrasesInCategory = presetsRepository.getPhrasesForCategory(category.categoryId)
val phrase = phrasesInCategory.findLast { localizedResourceUtility.getTextFromPhrase(p) == phraseString }
```

My bad if that's not what you're trying to attempt here and I think it might return a nullable as well? Maybe there's a `findLastOrNull {  }`

app/src/main/java/com/willowtree/vocable/presets/adapter/AddToCategoryPickerViewModel.kt
```diff
@@ -0,0 +1,154 @@
...
+                        }
+                    }
+
+                    // remove the category from the saved phrases map
+                    savedPhrases.remove(category.categoryId)
+                }
+            }
+
+            liveShowPhraseDeleted.postValue(true)
+            delay(KeyboardViewModel.PHRASE_ADDED_DELAY)
```
luonglegs, line 113: I think this is an example of me not having enuff context on the project. Why we are intentionally adding a `delay` here? :o

jollygreenegiant: This is for a dialog popup - post a value of true to show the dialog, wait two seconds, then post a value of false to dismiss it.  Not something a decision I made, but it's used in a couple places.

app/src/main/java/com/willowtree/vocable/presets/adapter/CustomCategoryAdapter.kt
```diff
@@ -0,0 +1,72 @@
...
+        itemView.isInvisible = true
+        parent.post {
+            with(itemView) {
+                findViewById<View>(R.id.filler_view).minimumHeight = getMinHeight(parent)
+                isInvisible = false
+            }
+        }
+
+        return CustomCategoryViewHolder(itemView)
+    }
```
luonglegs, line 53: Wouldn't it be easier to just pass in `CustomCategorySwitchItemBinding` into the `CustomCategoryViewHolder ` class? You could handle all your initial view stuff in there too since you have access to the `binding` val.

luonglegs: So you could do something like
```
override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ViewHolder {
        val inflater = LayoutInflater.from(parent.context)
        val binding = CustomCategorySwitchItemBinding.inflate(inflater, parent, false)

       // do your init binding stuff here

        return ViewHolder(binding)
    }
....

// Then in your view holder you'd just have reference to your `binding` already to use for the `bind()` method.
```

app/src/main/java/com/willowtree/vocable/settings/EditCategoriesViewModel.kt
```diff
@@ -144,6 +144,10 @@ class EditCategoriesViewModel : BaseViewModel() {
         return localizedResourceUtility.getTextFromCategory(updatedCategory)
     }
 
+    fun getUpdatedCategory(category: Category): Category {
+        return overallCategories.first { it.categoryId == category.categoryId }
```
luonglegs, line 148: Can the `first {  }` method return a null? What happens if it can't find it at all? The method signature says it's non-nullable so I figured there would be compiler errors here.

luonglegs (approved):
💰 Looks good to me! 💰 
I dropped a few questions and suggestions. No red flags or anything blocking tho.

app/src/main/java/com/willowtree/vocable/presets/adapter/AddToCategoryPickerViewModel.kt
jollygreenegiant: I don't think there's a way to avoid this with the way the model is set up - there isn't a simple way to check if a phrase is in a category just by using the phrase string.  I also would like to avoid nested loops if possible, but in this case it's really unlikely that the user would have tons of custom categories with tons of custom phrases in each one, and since it's all in a background scope, I think it should be okay.  I'll definitely keep it in mind if we start to see performance issues though

app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
jollygreenegiant: It's a non-nullable argument to the Navigation component, so I don't think it can be null, or it wouldn't even get to this screen.

app/src/main/java/com/willowtree/vocable/presets/AddToCategoryPickerFragment.kt
jollygreenegiant: Lol nope.  I'll fix it

app/src/main/java/com/willowtree/vocable/presets/adapter/AddToCategoryPickerViewModel.kt
```diff
@@ -47,12 +49,22 @@ class AddToCategoryPickerViewModel : BaseViewModel() {
...
+    fun buildCategoryList(phraseString: String, categories: List<Category>): Map<Category, Boolean> {
         val phraseMap = mutableMapOf<Category, Boolean>()
 
-        backgroundScope.launch {
-            userCategories = presetsRepository.getUserGeneratedCategories()
+        categories.forEach {
+            phraseMap[it] = false
+        }
+
+        return runBlocking {
```
nicholas-cook, line 65: Why the switch to `runBlocking` here? This will block the UI thread for the user.

nicholas-cook (approved):

PR #289, opened 2020-09-04
Added some keyboard tests and one main screen test.

jennydoesthings:
**Added in Espresso tests**

- verifyDefaultSayingsInCategoriesExist: Main screen test to make sure that the General category is the first you see and holds the phrases we expect.
- Rearranged the MainScreenTest a bit to group the tests by functionality.
- Several tests for the keyboard screen testing the default view, that phrases show up, that you can remove letters, reset the phrase shown, and that you can navigate back to the home screen.

**Added in support for Espresso tests**
- Added in an assertElementExists function to ViewInteractionsExt.
- Added in the keyboard container to the MainScreen. 
- Described the keyboard screen.
- Added a function to let you type in one or more characters. It handles characters that we don't currently support.

**Known Issues**
-  I don't love the variable name I picked for the keyboard container but nothing else came to mind.
- Could add a little more error handling around the keyboard input function. 🤷‍♀️ 

**Next steps**
- This should set us up well to be able to work through custom categories and phrases with the keyboard input function.
- Another avenue to run down is Settings--there are currently no Settings tests which should be a pretty easy win.

### DIFF
--- a/app/src/androidTest/java/com/willowtree/vocable/screens/KeyboardScreen.kt
+++ b/app/src/androidTest/java/com/willowtree/vocable/screens/KeyboardScreen.kt
@@ -1,0 +1,34 @@
+package com.willowtree.vocable.screens
+
+import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.NoMatchingViewException
+import androidx.test.espresso.matcher.ViewMatchers.withId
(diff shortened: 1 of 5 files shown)


SashankPatel (approved):
:rocket:

jollygreenegiant (approved):

PR #295, opened 2020-09-09
Fix presets empty state for landscape, and remove My Sayings

### LINKED ISSUE #242: Remove "My Sayings" as Menu Item
**Acceptance Criteria:**

- "My Sayings" category should become a custom category
- The "Edit My Sayings" page should be removed from the Settings screen.

**Design Link (Home Screen):** 
https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable-New?node-id=19%3A0

**Design Link (Settings Screen):**
https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable-New?node-id=19%3A1

### LINKED ISSUE #288: No empty state for landscape orientation
Sept 3rd Bug 1: No empty state in Landscape view

Device: Pixel 2

Scenario:
1.) Hold the phone in Portrait View.
2.) Launch the app
3.) Add a New Category without phrases
4.) Navigate to the Home screen and select newly created category. -> Empty state appears.
5.) Rotate the phone in landscape view

Expected: 
Expected_EmptyState

Actual:
Actual_EmptyState

_Originally posted by @SashankPatel in https://github.com/willowtreeapps/vocable-android/pull/287#issuecomment-686448199_

jollygreenegiant:
Description: 
This was originally just to fix the empty state for landscape, but in doing so, created an issue where the empty state would show on the My Sayings tab as well.  My Sayings is due to be removed anyway, so I just did that here.  Technically, the My Sayings category is still in the database, I just changed the `getAllCategories()` call to exclude it, so that we don't have to do any database migrations.  I also removed it from the settings screen, per the requirements of #242.

Closes #288 
Closes #242 

- [ ] Regression Testing


SashankPatel:
Good to go from my end: 
Verified empty state appears in portrait and landscape (pixel phone)
Verified a phrase can be entered via the Add phrase empty state button.
- Verified My Sayings was removed from Home screen, Settings screen, and Categories view.

nicholas-cook (commented):
Are we keeping the user's existing phrases from "My Sayings" in the app? Wondering if we need to delete that category from the database or if they can still access it from the category settings screen.

app/src/main/java/com/willowtree/vocable/splash/SplashViewModel.kt
```diff
@@ -21,7 +26,51 @@ class SplashViewModel : BaseViewModel() {
...
+            }
+
+            // Create a new custom category called "My Sayings" that is user-generated
+            val newCategoryId = UUID.randomUUID().toString()
+            val newCategory = Category(
+                newCategoryId,
+                System.currentTimeMillis(),
+                true,
+                null,
+                mapOf(Pair(Locale.getDefault().toString(), "My Sayings")),
```
nicholas-cook, line 54: I would grab the localization pairs from the old "My Sayings" category so we can preserve language versions here.

app/src/main/java/com/willowtree/vocable/splash/SplashViewModel.kt
```diff
@@ -21,7 +26,51 @@ class SplashViewModel : BaseViewModel() {
...
+            presetsRepository.addCategory(newCategory)
+
+            // Get the phrases from the old My Sayings category and add cross refs with the new category
+            val mySayingsPhrases = presetsRepository.getPhrasesForCategory(PresetCategories.USER_FAVORITES.id)
+            mySayingsPhrases.forEach {
+                presetsRepository.addCrossRef(CategoryPhraseCrossRef(newCategoryId, it.phraseId))
+            }
+
+            // Get the old My Sayings category and delete it
+            val mySayingsCategory = presetsRepository.getCategoryById(PresetCategories.USER_FAVORITES.id)
```
nicholas-cook, line 68: Should we also delete the old category if the user didn't have any saved phrases? That way it's not hanging around in memory.

jollygreenegiant: Yeah, that's what the next line is doing, right?  It gets the old My Sayings category on line 68, then deletes it on line 69

app/src/main/java/com/willowtree/vocable/splash/SplashViewModel.kt
jollygreenegiant: Good catch 👍

app/src/main/java/com/willowtree/vocable/splash/SplashViewModel.kt
jollygreenegiant: jk I see what you're saying, you're right

app/src/main/java/com/willowtree/vocable/splash/SplashViewModel.kt
```diff
@@ -20,8 +29,83 @@ class SplashViewModel : BaseViewModel() {
 
     private fun populateDatabase() {
         backgroundScope.launch {
+            val newCategoryId = UUID.randomUUID().toString()
+            val shouldUpdateName = presetsRepository.getPhrasesForCategory(PresetCategories.USER_FAVORITES.id)
```
nicholas-cook, line 33: Do you need this val? Looks like it's not being used

app/src/main/java/com/willowtree/vocable/splash/SplashViewModel.kt
```diff
@@ -20,8 +29,83 @@ class SplashViewModel : BaseViewModel() {
...
+            presetsRepository.getCategoryById(PresetCategories.USER_FAVORITES.id)
+
+        // if the user has My Sayings phrases, we need to migrate them to a custom category
+        if (presetsRepository.getPhrasesForCategory(PresetCategories.USER_FAVORITES.id)
+                .isNotEmpty()
+        ) {
+            val allCategories = presetsRepository.getAllCategories()
+
+            // Get the index of the first hidden category to find the sort order of new category
+            var firstHiddenIndex = allCategories.indexOfFirst { it.hidden }
```
nicholas-cook, line 75: Could we also try to pull the sort order from the old My Sayings category first and then calculate it if it's null?

nicholas-cook (approved):
Just a couple of small comments, looks good!

app/src/main/java/com/willowtree/vocable/splash/SplashViewModel.kt
jollygreenegiant: I don't think we want to do that, because it's considered a custom category now, so it should be at the end, right?

nicholas-cook: Ah fair point

PR #313, opened 2020-09-17
Separate debug and release traffic for Firebase

### LINKED ISSUE #312: Create Build Variant that Does Not Include Crashlytics
Create build without crashlytics.

jollygreenegiant:
Description: 
Separates release and debug traffic to separate apps within the Firebase project.  This also removes `google-services.json` from the `.gitignore` file, because it is safe to have in there for open source use, and is required to be able to run the app now that Firebase is integrated.  The user could still manually run on a release variant if they wanted, which would send traffic to the Vocable Android project instead of Vocable Android Debug, but that will be recommended against in the documentation and shouldn't pose an issue.

Closes #312 

- [ ] Regression Testing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,9 +43,6 @@ captures/
 # External native build folder generated in Android Studio 2.2 and later
 .externalNativeBuild
 
-# Google Services (e.g. APIs or Firebase)
-google-services.json
-
 # Freeline
 freeline.py
 freeline/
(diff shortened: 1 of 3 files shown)


jennydoesthings (approved):
🐐 LGTM 🐐

nicholas-cook (approved):

PR #323, opened 2020-10-05
Crash fix and bump version

jollygreenegiant:
Description: 
This should fix [this crash from Crashlytics](https://console.firebase.google.com/u/0/project/vocable-fcb07/crashlytics/app/android:com.willowtree.vocable/issues/54cc380eaee140c588c8467990c327f2?time=last-seven-days&sessionEventKey=5F7A774500BE0001293C132E410742D6_1458388187849643302) that I've seen a couple of times in different places.  As far as I'm aware, there is no way to do this that's particularly elegant, so this should fix the crash without affecting regular use.

- [ ] Regression Testing

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -18,8 +18,8 @@ android {
         applicationId "com.willowtree.vocable"
         minSdkVersion 24
         targetSdkVersion 29
-        versionCode 26
-        versionName "1.4.1"
+        versionCode 27
+        versionName "1.5.0"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         javaCompileOptions {
             annotationProcessorOptions {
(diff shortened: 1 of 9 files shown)


CarolineLaw (commented):
looks like you had to add in a check to make sure that the current destination is the one that you're in, think you can help me understand how this fixes things? I'm sure it does, I'm just wondering :] other than that things looking good to me :D I'd ask someone else to also take a look though!

jollygreenegiant:
@CarolineLaw this is the error from Crashlytics:
Screen Shot 2020-10-06 at 9 43 56 AM

Basically the way I've seen this caused before in other apps is that a user clicks twice really quickly on a button (or somehow a double tap is registered).  The first tap changes the Navigations currentDestination, and the second tap happens before that process is able to be completed, so if you're trying to go from Fragment A to Fragment B, on the second tap, it thinks you're already on B, so it's trying to go from A to B while it's already on B, so it crashes.  So I just added a check to each navigation that says "if you're on Fragment A, go from A to B", to avoid that crash.

CarolineLaw (approved):
👨🏼‍🍳 💋

jennydoesthings (approved):
🐐 LGTM 🐐

PR #336, opened 2020-10-22
Reverting schema 4 to fix migration tests

jennydoesthings:
Schema 4 was overwritten with schema 5 which incorrectly gave it the timestamp column.  This PR removes that column, allowing the tests to proceed correctly.

I also added some comments as to the reason we have these tests in place.


nerdsosky (approved):
Looks good to me 👍

PR #337, opened 2020-10-23
Adding a 4to5 migration test.

jennydoesthings:
Added a test that looks at the migration from schema 4 to schema 5 (adding a timestamp.)

I based this off of one of the other tests in this file--it should create a phrase, migrate the database, then confirm the phrase is returned correctly.  This is my first go at writing a test like this, so feedback is much appreciated.

### DIFF
--- a/app/src/androidTest/java/com/willowtree/vocable/room/MigrationTest.kt
+++ b/app/src/androidTest/java/com/willowtree/vocable/room/MigrationTest.kt
@@ -184,4 +184,59 @@ class MigrationTest {
                 close()
             }
     }
+
+    @Test
+    @Throws(IOException::class)
+    fun migrate4to5() {
+        // This migration adds a timestamp to phrases.
+
+        val categoryNameMap = HashMap<String, String>()
+        categoryNameMap["en_US"] = "My Sayings"
+
+        val phraseOneMap = HashMap<String, String>()
+        phraseOneMap["en_US"] = "Saying 1"
+
+        val categoryNameV4 = Converters.stringMapToJson(categoryNameMap)
+        val testPhraseV4 = Converters.stringMapToJson(phraseOneMap)
+
+        val phraseId = UUID.randomUUID().toString()
+        val creationDate = System.currentTimeMillis()
+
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/com/willowtree/vocable/room/MigrationTest.kt
```diff
@@ -184,4 +184,59 @@ class MigrationTest {
...
+                crossRefCursor.close()
+
+                    val phraseCursor = query("SELECT localized_utterance FROM Phrase WHERE phrase_id = '$phraseId'")
+                    while (phraseCursor.moveToNext()) {
+                        val saying = phraseCursor.getString(phraseCursor.getColumnIndex("localized_utterance"))
+                        phrasesAdded.add(saying)
+                    }
+                    phraseCursor.close()
+
+                Assert.assertTrue(phrasesAdded.contains(testPhraseV4))
```
nicholas-cook, line 237: We also need to assert that the `CategoryPhraseCrossRef` now has a `timestamp` field. So we'll want to pull that out of the database and check that it has the default value for that field (which I think is `null`). So we'd want something like:
`Assert.assertNull(categoryPhraseCrossRef.timestamp)`

app/src/androidTest/java/com/willowtree/vocable/room/MigrationTest.kt
```diff
@@ -235,6 +237,7 @@ class MigrationTest {
                     phraseCursor.close()
 
                 Assert.assertTrue(phrasesAdded.contains(testPhraseV4))
+                Assert.assertEquals(0L, timestamp)
```
CarolineLaw, line 240: v cool that the timestamp is initialized at 0L instead of null

jollygreenegiant: Yeah it's nullable for Kotlin reasons, because the db can return null, even if the field is non-nullable.  So I always make them nullable when I can so Kotlin forces you to handle it.

jollygreenegiant (approved):

PR #339, opened 2020-10-26
Unit tests PR (1/3) - set up base for unit tests

jollygreenegiant:
Description: 
This sets up a base for future unit tests to be written.  Rather than submit one enormous PR, I'm submitting this one, and then will make two more, one to finish the Koin refactor, and one (or several) more to make more unit tests.

**Koin refactor:** 

Previously, we were injecting a lot of dependencies as class fields, rather than through the constructor, which made those classes very difficult to test.  In this PR, I changed `PresetsViewModel` to have `PresetsRepository` injected through the constructor, so it can be easily mocked.  

Koin also provides an API for ViewModel injection, which I switched to for the `PresetsViewModel` here, and will switch to for the others as well, as part of a different PR that is specifically dedicated to that.

**Unit Tests:**
`PresetsViewModel` is the only class being tested in this PR (and it isn't complete).  To avoid this being a gigantic PR, I'll write tests for other classes, including the rest of the ViewModels and the repository, in a separate PR.


app/src/test/java/com/willowtree/vocable/screens/PresetsViewModelTest.kt
```diff
@@ -0,0 +1,131 @@
...
+    private lateinit var categoryObserver: Observer<Category>
+
+    @Mock
+    private lateinit var categoryListObserver: Observer<List<Category>>
+
+    @Mock
+    private lateinit var phraseListObserver: Observer<List<Phrase>>
+
+    @Mock
+    private lateinit var presetsRepository: PresetsRepository
```
nicholas-cook, line 46: I think we may want to move these to an interface pattern eventually. So we can have the "real" implementation that we use in the app and a "fake" one we use for tests. That way we don't have to use Mockito to mock responses.

PR #340, opened 2020-10-27
Unit tests PR (2/3) - Koin refactor

jollygreenegiant:
Description: 
This branch is built off of #339, and this PR should be merged into that branch before that branch goes into develop.

This refactors the way we use Koin so that ViewModel dependencies are now injected into the constructor as opposed to as class fields, which will significantly improve ease of testing.

- [ ] Regression Testing


app/src/main/java/com/willowtree/vocable/keyboard/KeyboardFragment.kt
```diff
@@ -140,10 +140,6 @@ class KeyboardFragment : BaseFragment<FragmentKeyboardBinding>() {
 
         (binding.phraseSavedView.root as TextView).setText(R.string.saved_successfully)
 
-        viewModel = ViewModelProviders.of(
```
CarolineLaw, line 143: this is probably better bc I think `ViewModelProviders.of` is deprecated now haha

nicholas-cook (approved):
Nice, this is a lot cleaner

PR #348, opened 2020-12-24
Fix custom categories and added back My Sayings

CarolineLaw:
Description: 
This is a huge PR I'm so sorry 😭 and it's not done 😞 
- Separates custom categories out so that it's categories and then their phrases instead of it being on one page
- Put the My Sayings category back and makes it so that when the user saves a phrase it gets added there

**!!!! Recents category DOES NOT work as expected:** when the app is closed only user generated phrases are saved to the Recents category. I will be making a new PR that fixes this by changing how we add things to the database. Right now we are deleting everything and re-adding them. I think updating phrases and categories instead would be better.

**ALSO:** I need design assistance with custom categories, I followed what was on iOS, but it might actually be different for Android. Also I need to know what to do when our Category Names are very large for portrait and landscape phones.

~also I need to fix testing :<~ Fixed testing, they pass now!!

Screenshots: 
Screen Shot 2020-12-24 at 11 06 12 AM
Screen Shot 2020-12-24 at 11 06 21 AM
Screen Shot 2020-12-24 at 11 06 30 AM
Screen Shot 2020-12-24 at 11 10 58 AM
Screen Shot 2020-12-24 at 11 07 54 AM

- [ ] Regression Testing


CarolineLaw:
The work here is done after I spoke to many designers. We discussed custom categories and how phrases are added to categories from the keyboard. After our discussion I decided this work was the best course of action. We will need to discuss when/if we want to merge this into develop and release this work.

We also need to do the work to change the way we add categories and phrases to the database because right now we are just deleting and re-adding, which is fine, but kind of lazy. It's also messing with how the Recents category works -- right now it clears out any non-user generated phrases (preset phrases). 

Please contact me at carolinealaw@gmail.com if you have any questions or concerns or just wanna say hi!

It's been a pleasure and an honor working at WillowTree with so many talented and smart people. I will miss you all dearly.

CarolineLaw:
@cxw320 @willdenne want to take a look at this branch? ;D if you like this work maybe we can work on this together and maybe make this a real pr haha!

CarolineLaw:
Seems the work on this branch is being done in other branches :D this is v cool and good and I'm happy to close this out 😊

PR #353, opened 2022-04-06
Bugfix/Phrases Update On Category Change

cxw320:
**Description:** 
In the iOS Vocable app, the list of phrases updates when a user changes the category via the forward and backward buttons.

In the Android Vocable app, this list of phrases does not update automatically when the user is scrolling through the category names. A user must click the category view to refresh the list of phrases.

In matching iOS and in the spirit of enabling the user a faster search for a specific phrase, this PR enables the list of phrases to update and match the category as it changes. It should reduce one more click for the user in the path to finding a phrase. 

Additionally, the "onClick" action for the CategoryButton is now removed since the user no longer needs to click on the category name to change the category. This is also consistent with the behavior in iOS. 

**Screenshots:** (sorry for the chonky screen recording, wasn't letting clip the god dang recording area)

Recording of **_old behavior_** (Phrases do not automatically update):

https://user-images.githubusercontent.com/14029158/162018760-fedd51e4-d1d6-48f1-bb7b-c67e98d63eb9.mov

Recording of **_new behavior_** (Phrases update with category scroll):

https://user-images.githubusercontent.com/14029158/162018487-c22e17a3-10d3-472c-9184-5d30465abb41.mov

- [ ] Regression Testing


app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
```diff
@@ -45,9 +45,13 @@ class PresetsFragment : BaseFragment<FragmentPresetsBinding>() {
             when (val currentPosition = binding.categoryView.currentItem) {
                 categoriesAdapter.itemCount - 1 -> {
                     binding.categoryView.setCurrentItem(0, true)
+                    val currentCategory = categoriesAdapter.getCategory( 0)
+                    presetsViewModel.onCategorySelected(currentCategory)
                 }
                 else -> {
                     binding.categoryView.setCurrentItem(currentPosition + 1, true)
+                    val currentCategory = categoriesAdapter.getCategory(currentPosition+1)
+                    presetsViewModel.onCategorySelected(currentCategory)
```
cxw320, line 54: The magic is happening Inside the binding.categoryForwardButton and the binding.categoryBackButton. *sparkle sparkle* 

When these are clicked, I'm extracting the current category via a new method I created "getCategory" that uses the currentPosition Int to index the list of categories from the "items" collection found in the CategoriesPagerAdapter.

I logged - er I mean - I timber'd (ofc) and printed the currentPosition as I clicked the forward button to change categories - and I noticed the currentPosition keeps growing to a very high number - it just increments the number 4ever...so in the getCategory method, i am just going to modulo that bad boi to get a correct index.

app/src/main/java/com/willowtree/vocable/presets/CategoriesFragment.kt
```diff
@@ -59,9 +59,6 @@ class CategoriesFragment : BaseFragment<CategoriesFragmentBinding>() {
             with(categoryButton.root as CategoryButton) {
                 tag = category
                 text = localizedResourceUtility.getTextFromCategory(category)
-                action = {
```
cxw320, line 62: If we leave this in and the user taps on the category title, there will be a quick glitch of the phrases as it "refreshes" again and it looks a bit strange for the user. Happy to leave this in or comment it out though if we don't want to completely delete it out.

willdenne (approved):
Looks good to me

app/src/main/java/com/willowtree/vocable/presets/CategoriesFragment.kt
CarolineLaw: Do we know why this was put in? Why we decided that we'd refresh the phrases when the user taps on the button? You can remove this if others thinks it's also okay 👍🏼 :3 but we should figure out how come we even put it in in the first place ;D

rhsalas3:
@cxw320 

I tested this branch and found some issues. I'll leave it up to your team if you decide they need to be fixed for this PR or not. But the issues involve the selected categories on the main screen and the expected phrases to appear.

Let me know if you have questions, etc.

There are 3 scenarios I am seeing unexpected behavior in:
 - [ ] Leaving the main screen and returning always "deselects" the category you were on when you navigated away from the screen. In addition to the category not being selected, the phrases that are available are always the preset phrases for the 'General' category. Leaving again and coming back to the main screen will update the screen to show/select the 'General' category, so that it is corrected.


https://user-images.githubusercontent.com/34660497/162497656-9c17142e-0572-464d-90fd-38560c0dc905.mov




- [ ] While in the state described above, clicking the category shows the category as selected but the phrases do not update to the correct phrases for that category.

Clicking_presentCategory_doesNotUpdate_phrases



- [ ] In the case of the '123' keypad category and an empty custom category, this state results in an additional UI error whereby the pagination label is missing.

VocableAndroid_returningToMainScreen_resultsIn_wrongCategoryAndPhrasesSelected

app/src/main/java/com/willowtree/vocable/presets/CategoriesFragment.kt
cxw320: Rudy (and Spud) pointed out that the tablet experience is quite different in that there are multiple categories visible at a time when a user scrolls forward/backward through the categories. Rudy pointed out that in iOS, the tablet and landscape experience requires the user to click on the category for the phrases to update.

Soooo...I put back this action callback so that the category titles can function as buttons 👍  :']

app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
```diff
@@ -110,6 +133,13 @@ class PresetsFragment : BaseFragment<FragmentPresetsBinding>() {
         binding.categoryView.registerOnPageChangeCallback(object :
             ViewPager2.OnPageChangeCallback() {
             override fun onPageSelected(position: Int) {
+
+                binding.phrasesPageNumber.text = getString(
+                        R.string.phrases_page_number,
+                        1,
+                        1
+                    )
```
cxw320, line 141: Having this here resolves the issue of the "Page 1 of 1" not appearing for the Recents and 123 category.

app/src/main/java/com/willowtree/vocable/presets/PresetsFragment.kt
```diff
@@ -29,6 +31,8 @@ class PresetsFragment : BaseFragment<FragmentPresetsBinding>() {
 
     private var maxCategories = 1
     private var maxPhrases = 1
+    private var isPortraitMode = true
+    private var isTabletMode = false
```
cxw320, line 35: Let me know if anyone objects to creating local variables in the PresetsFragment here for determining portrait mode / tablet mode...if in non-tablet mode and portrait mode, the phrases will automatically update with category forward and backward button.

rhsalas3:
@cxw320 

Caroline, I re-tested the branch with fixes for the issues I found. In my opinion, things looks good to go!

I did find other issues within the app, but not related to what you're working on in this PR. I am going to write those up as issues for the backlog...just fyi.

rhsalas3 (approved):
Tested and verified the fix on this PR's branch.

PR #354, opened 2022-04-12
Database rewrite

willdenne:
Description: 
This isn't a complete rewrite, but does remove the crossref table since we don't really need it. Instead, each phrase now has the id of the parent category. This makes dealing with the recents a bit more tedious, but overall I feel this change will make working with the database a lot easier. The database is now only created on the first time, instead of every time the user opens the app, and recents are no longer deleted every time.
Also added #360 which is to remove the is user generated so we can edit every category except recents.

Also:
Added Timber to replace log, and renamed USER_FAVORITES to MY_SAYINGS to better reflect whats going on.

We'll need to make sure going from the previous version to this version works fine, along with a regression.

- [ ] Regression Testing


willdenne:
Thinking I will https://github.com/willowtreeapps/vocable-android/issues/360 in this as well so we only have to write one migration

cxw320:
Not a code owner so I can't approve this - but I tested this PR and it looks good to me. (Found a regression that was caused by my PR though so I'll make a ticket for that)

PR #373, opened 2022-05-10
Remove button not appearing on categories

willdenne:
Quick regression PR, remove category should be not show on recents but should show for other categories.


cxw320 (approved):

CananArikan:
I am able to delete all categories including `Recents`. At the iOS version, we are able to remove Recents, only `My Sayings` `Remove category` button is disabled as expected. 
AllCategoriesAreDeleted

CananArikan:
Tested and verified the fix on this PR's branch.

PR #377, opened 2022-05-16
Fixed bug where the starting category would not be selected initially

willdenne:
Description: 
Fixed a bug where the General category was not selected initially.
For some reason `presetsViewModel.selectedCategory.value`'s data was stuck in the mPendingData while the actual data was null. I made it set from an observer instead.
- [x] Regression Testing


cxw320 (approved):

rhsalas3:
I was not able to reproduce the issue on develop with my emulator for an Android 12 (API 31) device. But testing off of this branch, I did not not find any issue where the General category was not the initial category selected. Lite regression didn't find any unknown issues. So this fix looks good to go to me.

PR #378, opened 2022-05-18
Back when save new text and toast

willdenne:
Description: 
Moves to the previous screen when saving a new category or phrase. Replaces the custom view with a toast, which makes persistence across screens easier. Note the screencap shows an android icon, should be a vocable one but I think that's an android bug.

Screenshots: 

- [x] Regression Testing


app/src/main/java/com/willowtree/vocable/settings/AddPhraseViewModel.kt
```diff
@@ -23,22 +24,29 @@ class AddPhraseViewModel : BaseViewModel() {
...
-                    0L,
-                    categoryId,
-                    System.currentTimeMillis(),
-                    System.currentTimeMillis(),
-                    mapOf(Pair(Locale.getDefault().toString(), phraseStr)),
-                    mySayingsPhrases.size
+            if (mySayingsPhrases.none {
+                    it.localizedUtterance?.containsValue(phraseStr) == true
+            }) {
+                Timber.d("WILL: saving like normal")
```
cxw320, line 34: Don't forget to remove these Timbers!

app/src/main/java/com/willowtree/vocable/settings/AddPhraseViewModel.kt
```diff
@@ -23,22 +24,29 @@ class AddPhraseViewModel : BaseViewModel() {
...
+                    )
                 )
-            )
-
-            liveShowPhraseAdded.postValue(true)
-            delay(PHRASE_UPDATED_DELAY)
-            liveShowPhraseAdded.postValue(false)
+                liveShowPhraseAdded.postValue(true)
+            } else {
+                Timber.d("WILL: not saving")
```
cxw320, line 47: And here

app/src/main/java/com/willowtree/vocable/settings/AddPhraseViewModel.kt
willdenne: :0

cxw320 (approved):
I flagged a few Timber statements to remove. This is a much better user experience!

rhsalas3:
Tested and verified that when saving a new phrase, making edits to phrases, making edits to existing categories, and saving new categories the "Changes saved" message appears and the user is immediately returned to the previous screen from the keyboard.

Regarding the messages appearing as Android notifications instead of the expected Toast style message, I have noted this behavior and will work to write up a bug ticket to further investigate what caused the regression. It could be, as you said, an Android bug.

I tested on a tablet and phone emulated device running Android 12 (API 31).

willdenne:
@rhsalas3 I'm getting that Android droid icon on all of my other apps, [might be fixed by a device restart?](https://issuetracker.google.com/issues/202863198?pli=1)? Either way I don't know if we can do much else.

PR #379, opened 2022-05-19
Add "+ add phrase" button

willdenne:
Description: 
Adds a "+ Add Phrase" to the list of phrases in the PresetsFragment. User should not be able to add the the num keypad and recents categories.
Refactored some of the buttons to `NoSayTextButton` to keep things the same.
Screenshots: 

- [x] Regression Testing


app/src/main/java/com/willowtree/vocable/presets/adapter/PhraseAdapter.kt
```diff
@@ -6,23 +6,45 @@ import android.view.ViewGroup
...
+    RecyclerView.Adapter<PhraseAdapter.PhraseViewHolder>(), KoinComponent {
 
-    inner class PhraseItemViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
-        private val binding = PhraseButtonBinding.bind(itemView)
+    abstract inner class PhraseViewHolder(
+        itemView: View
+    ) : RecyclerView.ViewHolder(itemView) {
+        abstract fun bind(text: String, position: Int)
+    }
+
```
cxw320, line 30: Why do we need to create this abstract class?

willdenne: Ohh good catch, was trying another approach and gave up.

willdenne: Changing my answer here, we need that since we have 2 different ViewHolders, one for the regular button and the other for the add phrase button. They both have different views so they need different viewHolders.

cxw320 (approved):
Love it!!!

CananArikan:
Thanks for this update! Test verified! All preset and custom categories have `+Add Phrase` button! 

Test Device: Phone Emulator Android 12 (API 32)
Version: V 1.5.1

PR #403, opened 2023-08-22
Add workflow to automate Google Play releases on semantic version tag

### LINKED ISSUE #399: Deploy builds from main to Google Play Production on tag
Given: an engineer is contributing code
When: they push up a tag that matches semantic versioning eg. "1.6.0"
Then: automation should build and test the app, supplying the version name from the tag, create a GitHub release, and deploy the artifact to Google Play

Note:
We will want to enable tag protection rules with this, remember this is a public repo!

PaulKlauser:
Fixes #399

### DIFF
--- a/.github/workflows/ps-release.yml
+++ b/.github/workflows/ps-release.yml
@@ -1,0 +1,60 @@
+name: Build & Release APK to PlayStore
+
+on:
+  push:
+    tags:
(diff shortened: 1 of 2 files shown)


SashankPatel (approved):
Awesomesauce!!

Issue #408, opened 2023-08-29
Make Recents category a query, not an entry in the Category table


PaulKlauser:
Currently the categories in `PresetCategories` are all created in the Category table. This creates a problem because the Recents category should really just be populated as a query on the Phrase table, not a separate category. This will resolve potential migration issues, as well as confusion around duplication of Phrases when added to recents.

TODO: 
* AC around recents behavior
* Does "My Sayings" relate to this as well?

PR #414, opened 2023-09-01
Chore/408 make recents a query 1

PaulKlauser:
Description: 
First steps in separating our Category domain model from the database model. Renamed the DB model and stopped duplicating Category state in PresetsViewModel.

Recommend stepping through by-commit. PR is a little larger than I wanted, but I wanted to get the tests in for PresetsViewModel to add confidence in the review and hopefully make it a little easier.

Part of #408

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
+++ b/app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
@@ -1,11 +1,11 @@
 package com.willowtree.vocable
 
-import com.willowtree.vocable.presets.PresetsRepository
+import com.willowtree.vocable.presets.IPresetsRepository
 import com.willowtree.vocable.room.CategoryDto
 import kotlinx.coroutines.flow.Flow
 
 class CategoriesUseCase(
-    private val presetsRepository: PresetsRepository
+    private val presetsRepository: IPresetsRepository
 ) {
 
     fun categories(): Flow<List<CategoryDto>> {
--- a/app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
+++ b/app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
@@ -1,0 +1,15 @@
+package com.willowtree.vocable
+
+import com.willowtree.vocable.presets.PresetsRepository
+import com.willowtree.vocable.room.CategoryDto
+import kotlinx.coroutines.flow.Flow
+
+class CategoriesUseCase(
+    private val presetsRepository: PresetsRepository
+) {
+
+    fun categories(): Flow<List<CategoryDto>> {
+        return presetsRepository.getAllCategoriesFlow()
+    }
+
+}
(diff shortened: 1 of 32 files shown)


app/src/main/java/com/willowtree/vocable/presets/PresetsRepository.kt
```diff
@@ -1,34 +1,40 @@
...
+    override suspend fun getPhrasesForCategory(categoryId: String): List<Phrase> {
         return database.categoryDao().getCategoryWithPhrases(categoryId)?.phrases ?: listOf()
     }
 
     suspend fun addPhrase(phrase: Phrase) {
         database.phraseDao().insertPhrase(phrase)
     }
 
-    suspend fun addPhraseToRecents(phrase: Phrase) {
+    override suspend fun addPhraseToRecents(phrase: Phrase) {
```
CarolineLaw, line 37: do you still need to add phrase to recents if you changed he way we get them?

app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import com.willowtree.vocable.presets.IPresetsRepository
+import com.willowtree.vocable.room.CategoryDto
+import kotlinx.coroutines.flow.Flow
+
+class CategoriesUseCase(
+    private val presetsRepository: IPresetsRepository
+) {
+
+    fun categories(): Flow<List<CategoryDto>> {
```
ryanbrookepayne, line 11: Any reason you aren't just invoking this use case given that there's only one function.
Ref: https://chrynan.codes/invoking-usecases-the-kotlin-way/

```suggestion
    operator fun invoke: Flow<List<CategoryDto>> {
```

app/src/main/java/com/willowtree/vocable/presets/PresetsRepository.kt
PaulKlauser: Good question. I haven't changed the way we get them yet, I've only started differentiating the Category model that we store in the database (`CategoryDto`), from the Category domain model (`Category`)

PaulKlauser: Getting myself back up to speed here 😅 I haven't actually even introduced `Category` yet in this code, only started adding the seams to do so, in this case `CategoriesUseCase`, which will eventually transform the database model into the domain model

app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
PaulKlauser: Yea I've seen that pattern, do you feel strongly that it's more readable that way? I've always felt it's pretty 6-in-one-half-a-dozen-in-the-other, and it gets awkward as soon as you want to add a second function

ryanbrookepayne: I don't feel strongly about it. Just thought I'd point it out.

PR #440, opened 2023-11-14
VOC-406 Introduces LocalesWithText as a way to abstract away from Map<String,String>

IanCrossCD:
Description: 

- Introduces LocalesWithText as a way to abstract away from Map<String,String>. 
- Allows us to move away from a Util and use methods and extensions instead. 
- Organizes a few util package files into locale package.
- No new DB manipulation should be necessary since its transformed from a string in both scenarios.

- [X] Acceptance Criteria satisfied
- [ ] Regression Testing


IanCrossCD:
I want to move away from using LocaleString all together but Locale was not friendly to conversions in my 10 minute attempt. I could at least strip it from Domain and below if our models had separation :)

app/src/main/java/com/willowtree/vocable/room/Converters.kt
```diff
@@ -24,10 +24,30 @@ object Converters : KoinComponent {
...
+    fun languagesWithTextToStringMap(json: String?): LocalesWithText? {
+        return json?.let {
+            val type = Types.newParameterizedType(Map::class.java, String::class.java, String::class.java)
+            val adapter: JsonAdapter<Map<String, String>> = moshi.adapter(type)
+            adapter.fromJson(it)?.let { stringMap ->
+                LocalesWithText(stringMap)
+            }
+
+        }
+    }
```
PaulKlauser, line 52: Dang, Moshi doesn't play nice with typealiases out of the box?

app/src/main/java/com/willowtree/vocable/utils/locale/LocalesWithText.kt
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * Gets the string corresponding to the given locale matched against LocaleString.
+     * @param locale The locale to match against turned into String
+     */
+    private operator fun get(locale: Locale): String? = localesTextMap[locale.toString()]
+
+    /**
+     * Sets the string corresponding to the given localeString.
+     * @param localeString
+     * @param text
```
PaulKlauser, line 46: Is there value in including `@param` tags that have no further description?

PaulKlauser (approved):
LGTM Thanks!

Try to keep an eye out for ways to break up PRs that touch 42 files. That can be by having multiple commits within the PR, where the big "name change" commits are isolated, or spreading the work across multiple PRs 😁

PR #444, opened 2023-11-16
Re-enable ui tests

acosta434:
Re-enable tests to see if they run in current pipeline

Description: 

- Removed ignore from existing tests
- Update string to fix failing test
- Add step to dismiss ArCoredialog

Screenshots: 
n/a

- [ ] Regression Testing

### DIFF
--- a/app/src/androidTest/java/com/willowtree/vocable/tests/BaseTest.kt
+++ b/app/src/androidTest/java/com/willowtree/vocable/tests/BaseTest.kt
@@ -1,12 +1,12 @@
 package com.willowtree.vocable.tests
 
-import android.content.Intent
+import android.Manifest
 import androidx.test.espresso.IdlingPolicies
 import androidx.test.espresso.IdlingRegistry
 import androidx.test.espresso.matcher.ViewMatchers
 import androidx.test.ext.junit.rules.ActivityScenarioRule
 import androidx.test.platform.app.InstrumentationRegistry.getInstrumentation
-import androidx.test.rule.ActivityTestRule
+import androidx.test.rule.GrantPermissionRule
 import androidx.test.uiautomator.By
 import androidx.test.uiautomator.UiDevice
 import androidx.test.uiautomator.Until
@@ -19,7 +19,6 @@ import org.junit.Rule
 import org.junit.rules.TestName
 import timber.log.Timber
 import java.io.File
-import java.lang.NullPointerException
 import java.util.concurrent.TimeUnit
 
 open class BaseTest {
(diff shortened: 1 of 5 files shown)


app/src/androidTest/java/com/willowtree/vocable/tests/BaseTest.kt
```diff
@@ -77,6 +79,8 @@ open class BaseTest {
         // Catch the exception if the popup doesn't appear so we don't fail the tests
         try {
             device.wait(Until.findObject(By.text("CONTINUE")), 10000).click()
+            Thread.sleep(2000)
+            device.pressBack()
```
PaulKlauser, line 83: Do we need to press continue, and then back? I think we can just press back and the dialog will dismiss.

app/src/androidTest/java/com/willowtree/vocable/tests/BaseTest.kt
```diff
@@ -46,10 +48,10 @@ open class BaseTest {
         // Since the build machine gets wiped after every run we can check the file storage
         // of the emulator to determine if this is a first time launch
         // and dismiss the full screen immersive popup on android if it is
-        firstLaunch = getInstrumentation().targetContext.filesDir.listFiles().isEmpty()
+/*        firstLaunch = getInstrumentation().targetContext.filesDir.listFiles().isEmpty()
         if (firstLaunch) {
             takeScreenshot("InitialSetup")
-        }
+        }*/
```
PaulKlauser, line 54: Is this intentionally commented out? If so, we should delete the code. We can bring it back through git if we need to.

app/src/main/AndroidManifest.xml
```diff
@@ -6,6 +6,9 @@
     <uses-permission android:name="android.permission.CAMERA" />
 
     <uses-feature android:name="android.hardware.camera.ar" />
+    <uses-feature
+        android:name="android.hardware.camera"
+        android:required="false" />
```
PaulKlauser, line 11: Was this change required to get the tests running?

PaulKlauser (changes requested):
Looking good, see my comments. Can we squash these commits down as well?

app/src/androidTest/java/com/willowtree/vocable/tests/BaseTest.kt
acosta434: No we do not. Pressing click just adds complexity. Removing the 'click' action also gets rid of the need to wait in the next step. Thanks

app/src/androidTest/java/com/willowtree/vocable/tests/BaseTest.kt
acosta434: That was part of a different problem. Removed.

app/src/main/AndroidManifest.xml
acosta434: It is not required right now, removing it.

acosta434:
@PaulKlauser I got it down to the 1 commit.

PaulKlauser (approved):

PR #453, opened 2023-11-27
Implement adding a Category in CategoriesUseCase

PaulKlauser:
Description: 
Implement adding a Category in CategoriesUseCase

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
+++ b/app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
@@ -4,20 +4,17 @@ import com.willowtree.vocable.presets.Category
 import com.willowtree.vocable.presets.IPresetsRepository
 import com.willowtree.vocable.presets.PresetCategoriesRepository
 import com.willowtree.vocable.presets.asCategory
-import com.willowtree.vocable.room.CategoryDto
 import com.willowtree.vocable.room.CategorySortOrder
 import com.willowtree.vocable.room.StoredCategoriesRepository
-import com.willowtree.vocable.utils.DateProvider
 import com.willowtree.vocable.utils.UUIDProvider
-import com.willowtree.vocable.utils.locale.LocalesWithText
 import com.willowtree.vocable.utils.locale.LocaleProvider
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
```diff
@@ -49,10 +46,9 @@ class CategoriesUseCase(
     }
 
     override suspend fun addCategory(categoryName: String, sortOrder: Int) {
-        presetsRepository.addCategory(
```
IanCrossCD, line 52: Is the `PresetsRepository.addCategory` still used?

app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
```diff
@@ -49,10 +46,9 @@ class CategoriesUseCase(
     }
 
     override suspend fun addCategory(categoryName: String, sortOrder: Int) {
```
IanCrossCD, line 48: Separate discussions from your code but I'm curious on your thoughts on these sort of mono use cases.

app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
PaulKlauser: Great ask, looks like it's not

app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
PaulKlauser: Meaning, use cases that can do more than one thing? I don't have super strong feelings, and I know a lot of folks go with the one-operation use cases, but I find that doesn't buy you a whole lot for the extra verbosity. What are your thoughts?

IanCrossCD: It doesn't buy you much for sure. It feel like its conflating two patterns but doesn't feel too wrong otherwise.

IanCrossCD (approved):
LGTM

PR #454, opened 2023-11-29
Androidx activityResult refactor

IanCrossCD:
Description: 
Migrates FaceTrackingManager Camera Permissions to Androidx activityResult. Creates CameraPermissionHelper


app/src/main/java/com/willowtree/vocable/utils/FaceTrackingManager.kt
```diff
@@ -51,11 +48,20 @@ class FaceTrackingManager(
...
+
+            cameraPermissionsHelper.permissionCallback = { isGranted ->
+                if (isGranted) {
+                    faceTrackingPermissions.enableFaceTracking()
+                } else {
+                    faceTrackingPermissions.disableFaceTracking()
+                }
+            }
+
             CoroutineScope(Dispatchers.Main).launch {
```
PaulKlauser, line 60: I think I missed this in a previous PR, but this is dangerous since there is no bound to how long we're running this block. Could result in trying to hide and show things that aren't there if the owner of this instance (Activity/Fragment) is gone. 

My recommendation is to make `initialize()` suspend, and then whoever is calling initialize is running that coroutine within their own lifecycle-bound scope.

Relating it to the current work: I'm actually wondering if you can simplify the back-and-forth of having the permission callback, and then updating faceTrackingPermissions, and then observing that update to faceTrackingPermissions. Would abstracting over the callbacks with a coroutine be possible here to make the code read more synchronously?

app/src/main/res/values/strings.xml
```diff
@@ -179,8 +179,14 @@
...
     <string name="no_recent_phrases_message">Start using the app to see your most recently used phrases here.</string>
     <string name="categories_edit_title">Categories</string>
-    <!-- TODO: Remove or change this placeholder text -->
-    <string name="hello_blank_fragment">Hello blank fragment</string>
+
+    <!-- Camera Permissions -->
+    <string name="permissions_missing_dialog_title">Required Permissions Unavailable</string>
+    <string name="permissions_missing_dialog_body">Camera permissions are required to use the Head Tracking feature. Please enable them in the app settings.</string>
+    <string name="permissions_confirm">OK</string>
+    <string name="permissions_cancel">Cancel</string>
```
PaulKlauser, line 187: Looks like we already have "Cancel" in `settings_dialog_cancel`. Wondering if we just use the pre-existing "Continue" as well instead of adding "OK". Less translation-changes to deal with.

app/src/main/res/values/strings.xml
```diff
@@ -179,8 +179,14 @@
...
+    <string name="permissions_rationale_title">Camera permissions are needed to enable Head Tracking.</string>
```
PaulKlauser, line 188: Intentional that this title has a period, and `permissions_missing_dialog_title` does not?

app/src/main/java/com/willowtree/vocable/facetracking/FaceTrackFragment.kt
```diff
@@ -75,7 +74,7 @@ class FaceTrackFragment : ArFragment() {
         ReplaceWith("Nothing, this is what AR is expecting")
     )
     override fun onRequestPermissionsResult(requestCode: Int, permissions: Array<out String>, grantResults: IntArray) {
-        EasyPermissions.onRequestPermissionsResult(requestCode, permissions, grantResults, requireActivity())
+        requireActivity().onRequestPermissionsResult(requestCode, permissions, grantResults)
```
PaulKlauser, line 77: I see this needs to be overridden because we're still extending from `ArFragment`, but can this be a no-op, since we're handling this via the ActivityResult stuff in MainActivity?

PaulKlauser (changes requested):
Sorry for the long lead-time here. Overall I like the changes, but I'm wondering if there's a way to simplify the ping-ponging between permissions callbacks and make that read more synchronously. There may not be, but worth poking at imo

app/src/main/java/com/willowtree/vocable/utils/FaceTrackingManager.kt
mattttvaughn: Re: our discussion offline, the dream of making permissions request work with suspend functions looks like an impossibility for now. Did some refactoring to move CameraPermissionsHelper func into FaceTrackingPermissions since they do the same thing, which eliminated the need for the callbacks.

app/src/main/java/com/willowtree/vocable/utils/FaceTrackingPermissions.kt
```diff
@@ -23,4 +43,82 @@ class FaceTrackingPermissions(private val sharedPreferences: IVocableSharedPrefe
...
+
+    private val permissionLauncher: ActivityResultLauncher<String> =
+        activity.registerForActivityResult(ActivityResultContracts.RequestPermission()) { isGranted: Boolean ->
+            if (isGranted) {
+                enableFaceTracking()
+            } else {
+                disableFaceTracking()
+                permissionSettingsDialog.show()
+            }
+        }
```
mattttvaughn, line 55: all the new stuff in this file is just moved wholesale from CameraPermissionsHelper. The only intended changes are to directly use `enableFaceTracking()` or `disableFaceTracking()` instead of calling them via callbacks

app/src/test/java/com/willowtree/vocable/utils/FaceTrackingPermissionsTest.kt
```diff
@@ -1,19 +1,36 @@
...
+            ContextCompat.checkSelfPermission(any(), any())
+        } returns PackageManager.PERMISSION_DENIED
+    }
+
     private fun createFaceTrackingPermissions(sharedPreferences: IVocableSharedPreferences): IFaceTrackingPermissions {
-        return FaceTrackingPermissions(sharedPreferences)
+        return FaceTrackingPermissions(
+            sharedPreferences = sharedPreferences,
+            activity = mockk(relaxed = true)
+        )
```
mattttvaughn, line 33: holler if there's a better approach since I know relaxed mocks are probably a code smell, but an activity is required in this class now and doing interface shenanigans didn't feel worth the time/complexity to me

app/src/main/java/com/willowtree/vocable/utils/FaceTrackingPermissions.kt
PaulKlauser: Since enable/disable are now called internally, do they still need to be part of the public interface here?

This then feeds into how this class is being tested, which I think is going to inform whether or not your mocking below is sufficient. My gut says it's not, since you're now unable to test the various branching points at line 49

app/src/main/java/com/willowtree/vocable/utils/permissions/FakePermissionRegisterForLaunch.kt
PaulKlauser, line 1: Can these Fakes go in the test directory?

mattttvaughn: Moved!

PaulKlauser (approved):

Issue #452, opened 2023-11-27
Properly abstract over saved/preset Categories


PaulKlauser:
`CategoriesUseCase` as it stands is only partially delegating between stored and preset categories. This class needs to serve as an interface that the app can interact with to get and update categories without concern as to whether or not they were presets or custom.

The main operations supported here are:
* Creating a new category
* Getting the list of categories
* Getting a category by ID
* Updating categories
* ~Deleting categories~ moving to a separate ticket

These all need to be handled properly for preset and user-made categories.

From the user's perspective, this will look like being able to add, hide, and rename categories, with no differentiation from preset to custom. Updating the app will not remove user-made categories, or re-add presets they've removed. 

It may update an existing preset, in the case that it was not modified, and we found an issue with a translation, for example.

PR #458, opened 2023-12-28
Rename PresetsRepository

PaulKlauser:
The name already didn't match the responsibility of the class, and with the migration, the name was even more confusing.

#452

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/presets/ILegacyCategoriesAndPhrasesRepository.kt
+++ b/app/src/main/java/com/willowtree/vocable/presets/ILegacyCategoriesAndPhrasesRepository.kt
@@ -6,8 +6,9 @@ import com.willowtree.vocable.room.PhraseDto
 import com.willowtree.vocable.utils.locale.LocalesWithText
 import kotlinx.coroutines.flow.Flow
 
-//TODO: PK - Rename this once we make the jump to rename [PresetsRepository] -> "RoomPresetsRepository"
-interface IPresetsRepository {
+@Deprecated("This is the old way of accessing categories and phrases. Prefer using" +
+        " ICategoriesUseCase instead.")
+interface ILegacyCategoriesAndPhrasesRepository {
     suspend fun getPhrasesForCategory(categoryId: String): List<PhraseDto>
 
     /**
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/willowtree/vocable/presets/ILegacyCategoriesAndPhrasesRepository.kt
```diff
@@ -6,8 +6,9 @@ import com.willowtree.vocable.room.PhraseDto
 import com.willowtree.vocable.utils.locale.LocalesWithText
 import kotlinx.coroutines.flow.Flow
 
-//TODO: PK - Rename this once we make the jump to rename [PresetsRepository] -> "RoomPresetsRepository"
-interface IPresetsRepository {
+@Deprecated("This is the old way of accessing categories and phrases. Prefer using" +
+        " ICategoriesUseCase instead.")
```
mattttvaughn, line 10: Nit but would we want `PhrasesUseCase` mentioned in here as well?

mattttvaughn (approved):
Name change here is great 👍

Is the implication that we'll also eventually deprecate `StoredCategoriesRepository` and prefer exposing everything through the UseCases? Or was this guy just too insanely overloaded with stuff?

app/src/main/java/com/willowtree/vocable/presets/ILegacyCategoriesAndPhrasesRepository.kt
PaulKlauser: Good call, I forgot that existed

PaulKlauser:
> Name change here is great 👍
> 
> Is the implication that we'll also eventually deprecate `StoredCategoriesRepository` and prefer exposing everything through the UseCases? Or was this guy just too insanely overloaded with stuff?

`StoredCategoriesRepository` will still exist, but the UI will only interact via the UseCases, which will then delegate to the appropriate stored/preset repositories as necessary

app/src/main/java/com/willowtree/vocable/presets/ILegacyCategoriesAndPhrasesRepository.kt
PaulKlauser: I'ma fix this forward in a following PR 👍

PR #460, opened 2024-01-03
Handle updating category name regardless of whether it's preset or stored

PaulKlauser:
#452

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
+++ b/app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
@@ -29,7 +29,7 @@ class CategoriesUseCase(
 
     override suspend fun getCategoryById(categoryId: String): Category {
         return storedCategoriesRepository.getCategoryById(categoryId)?.asCategory()
-            ?: presetCategoriesRepository.getCategoryById(categoryId)
+            ?: presetCategoriesRepository.getCategoryById(categoryId)?.let { if (it.hidden) null else it }
             ?: throw IllegalArgumentException("Category with id $categoryId not found")
     }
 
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
```diff
@@ -29,15 +29,40 @@ class CategoriesUseCase(
...
+                        presetCategory.categoryId,
+                        localizedName,
+                        presetCategory.hidden,
+                        presetCategory.sortOrder
+                    )
+                )
+            } else {
+                throw IllegalArgumentException("Category with id $categoryId not found")
+            }
+        }
```
mattttvaughn, line 65: Just curious- is this a solved problem? Throw exception vs. return result (vs silently swallow/log errors 👀)

If you've got any sage words or content you could point me to about this I'd appreciate it. Otherwise I can just research it a little and discuss with the smart folks at office hours tomorrow

mattttvaughn (approved):

app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
PaulKlauser: My stance on exceptions [is generally aligned with this guy](https://james1345.medium.com/exceptions-result-types-and-kotlin-235af74e43e0).

In this case, this method being called with a category ID that cannot be found is *not* an expected state of our application. It could be argued that we don't want the app to crash in this case, but I'd also argue that there's a benefit to a crash in that it's very obvious something has gone wrong, and requires the least lift _right now_.

I think at the least we'd want a result return type, so that we could communicate to the caller that things broke, so that it can react accordingly. If we fail silently, we may leave the user stuck in an invalid state, which is worse than a crash IMO. 

Granted, as I type this, perhaps a crash in Vocable _is_ worst case, because the user may not be able to launch Vocable again themselves if they have limited mobility.

What are your thoughts?

Issue #401, opened 2023-07-24
Don't require camera permission if head tracking is disabled


PaulKlauser:
**Scenario 1: Head Tracking Disabled**
Given: I do not have head tracking enabled
When: I launch the app
Then: I should not see a prompt for the camera permission
And: The app should function as normal

**Scenario 2: Enabling Head Tracking**
Given: Head tracking has never been enabled and camera permission haven't been granted
When: I enable head tracking
Then: I should be prompted for the camera permission

Note: Declining permission is not being explicitly handled as part of this ticket. That can be follow-on work.

The underlying issue here is that we're using an out of the box Fragment for the head tracking that comes from mlkit. This Fragment handles all of the permission requesting, which is great, but it also means we don't have as much flexibility with this handling.

PaulKlauser:
This is likely going to require migrating away from extending ArFragment as a whole, and instead just including the SceneView. This will also allow us to avoid the app closing on permission denial.

PR #468, opened 2024-01-10
Test Face Tracking Permission Dialogs, Remove PermissionRequested

mattttvaughn:
Description: 
Adds more testing around FaceTrackingPermissionsTest, covering the dialog behavior and a few permission acceptance/denial cases. Also refactors out the FaceTrackingPermission permissions checks so we don't need to mock activity/ActivityCompat/ContextCompat in the tests anymore.

Removes PermissionsRequested, as that was no longer being consumed by the UI.

Screenshots: 
N/A

#401


app/src/main/java/com/willowtree/vocable/utils/permissions/ActivityPermissionsChecker.kt
```diff
@@ -0,0 +1,26 @@
...
+class ActivityPermissionsChecker(
+    private val activity: AppCompatActivity,
+) : PermissionsChecker {
+
+    override fun hasPermissions(permission: String): Boolean {
+        return ActivityCompat.checkSelfPermission(
+            activity,
+            Manifest.permission.CAMERA
+        ) == PackageManager.PERMISSION_GRANTED
+    }
```
PaulKlauser, line 17: Doesn't look like you actually used the `permission` parameter here

app/src/main/java/com/willowtree/vocable/utils/permissions/ActivityPermissionsChecker.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    override fun shouldShowRequestPermissionRationale(permission: String): Boolean {
+        return ActivityCompat.shouldShowRequestPermissionRationale(
+            activity,
+            Manifest.permission.CAMERA
+        )
+    }
```
PaulKlauser, line 24: Ditto

app/src/test/java/com/willowtree/vocable/utils/FaceTrackingPermissionsTest.kt
```diff
@@ -66,60 +61,155 @@ class FaceTrackingPermissionsTest {
...
+        val permissions = createFaceTrackingPermissions(
+            sharedPreferences = sharedPreferences,
+            shouldShowRequestPermissionRationale = true,
+        )
         advanceUntilIdle()
 
+        // Launch initial rationale dialog, deny permissions
         permissions.requestFaceTracking()
-        assertEquals(permissions.permissionState.first(), IFaceTrackingPermissions.PermissionState.PermissionRequested)
+        fakePermissionsDialogShower.permissionDialogOnPositiveClick.invoke()
```
PaulKlauser, line 187: Shouldn't this be `onNegativeClick` if we're denying?

app/src/test/java/com/willowtree/vocable/utils/FaceTrackingPermissionsTest.kt
```diff
@@ -66,60 +61,155 @@ class FaceTrackingPermissionsTest {
...
+            assertEquals(0, fakePermissionsDialogShower.showSettingsPermissionDialogCalledCount)
+            fakePermissionsDialogShower.permissionDialogOnNegativeClick.invoke()
+            assertFalse(
+                "Head tracking should be disabled if user denies dialog",
+                sharedPreferences.getHeadTrackingEnabled()
+            )
+        }
+
+    @Test
+    fun `given initial rationale dialog denied, settings dialog rationale shown and updates head tracking when enabled`() = runTest {
```
PaulKlauser, line 177: This is a magilla of a test, but at the same time, I don't see a whole lot of other options for testing this fiddly permission stuff, and I like having these flows tested so 🤷

PaulKlauser (approved):
So can we remove mockk now? 😁

PaulKlauser (changes requested):
See comments

app/src/test/java/com/willowtree/vocable/utils/FaceTrackingPermissionsTest.kt
mattttvaughn: ah that's a failure of naming. "positive" click here means the user clicks okay to the rationale dialog so that the system permission dialog is shown. we deny via `fakePermissionRegisterForLaunch.triggerActivityResult`

Screenshot 2024-01-18 at 9 47 15 AM

I can make this more clear

app/src/test/java/com/willowtree/vocable/utils/FaceTrackingPermissionsTest.kt
mattttvaughn: I split out this test into two so it can at least verify head tracking permissions vs dialogs being shown separately. still doing a lot because of "fiddly permission stuff" reasons but should make it easier to track down reasons for test failure

app/src/test/java/com/willowtree/vocable/utils/FaceTrackingPermissionsTest.kt
mattttvaughn: I've updated `fakePermissionsDialogShower.permissionDialogOnPositiveClick` to `fakePermissionsRationaleDialogShower.rationaleDialogOnPositiveClick` to clarify this

PaulKlauser (approved):

PR #471, opened 2024-01-11
Refactor CategoriesUsCaseTest to use the actual DB

PaulKlauser:
The Fake was getting too complicated, and this is complex enough that I want the integration test. I've already found and fixed a few bugs.

#452


app/src/main/java/com/willowtree/vocable/presets/RoomPresetCategoriesRepository.kt
```diff
@@ -1,57 +1,82 @@
...
+            "Unknown preset category id: $categoryId"
+        )
+        val newPresetDto = PresetCategoryDto(categoryId, false, presetCategory.initialSortOrder)
+        database.presetCategoryDao().insertPresetCategory(newPresetDto)
+        Category.PresetCategory(
+            newPresetDto.categoryId,
+            newPresetDto.sortOrder,
+            newPresetDto.hidden,
+            presetCategory.getNameId()
+        )
```
mattttvaughn, line 56: Doesn't look like this `Category.PresetCategory` is used

app/src/main/java/com/willowtree/vocable/presets/RoomPresetCategoriesRepository.kt
```diff
@@ -1,57 +1,82 @@
...
+                Category.PresetCategory(
+                    it.categoryId,
+                    it.sortOrder,
+                    it.hidden,
+                    PresetCategories.values()
+                        .first { presetCategory -> presetCategory.id == it.categoryId }
+                        .getNameId()
+                )
+            }
+        }.onStart { ensurePopulated() }
```
mattttvaughn, line 33: nit: lots of nested `it`s

app/src/androidTest/java/com/willowtree/vocable/CategoriesUseCaseTest.kt
```diff
@@ -0,0 +1,353 @@
...
+            sortOrder = 0,
+            hidden = false,
+            resourceId = PresetCategories.GENERAL.getNameId()
+        ),
+        Category.PresetCategory(
+            categoryId = PresetCategories.BASIC_NEEDS.id,
+            sortOrder = 1,
+            hidden = false,
+            resourceId = PresetCategories.BASIC_NEEDS.getNameId()
+        ),
```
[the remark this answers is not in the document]
PaulKlauser: That's a good suggestion. Initially I was thinking that would kind of defeat the purpose of these tests, but this list is used to focus on the combination of preset and stored categories, so the specific content of the presets isn't relevant to the test. I'm in!

app/src/main/java/com/willowtree/vocable/presets/RoomPresetCategoriesRepository.kt
PaulKlauser: Good feedback. This is super interesting, because copilot wrote this, and I didn't do a good enough job reviewing it to see that!

PaulKlauser: Writing it by hand, I think it would've stood out to me more

mattttvaughn (approved):
One minor outstanding comment I left that you might've missed- dropping the approval on it regardless since its minor- feel free to fix at your discretion

app/src/main/java/com/willowtree/vocable/presets/RoomPresetCategoriesRepository.kt
PaulKlauser: You're absolutely right

PaulKlauser: I swear this isn't me making an excuse, but I've definitely been leveraging copilot more lately, and I need to be better about reviewing what it spits out. I've been leaning on "it passes the tests, so it's good", but that can be confusing if it's adding in nonsense code like this.

mattttvaughn: No you just need a second copilot to auto-review for nonsense code 😎

PR #477, opened 2024-01-17
Bugfix/448 crash on obfuscated builds

### LINKED ISSUE #448: Resolve ClassNotFoundException on Category
Seeing a number of ClassNotFoundExceptions coming through Firebase. These appear to be coming through when our app is submitted to Google Play and run through their automatic review.

Firebase link: https://console.firebase.google.com/u/0/project/vocable-fcb07/crashlytics/app/android:com.willowtree.vocable/issues/15cc2b91cbd419815de535248e5bbd97?time=last-seven-days&types=crash&sessionEventKey=655E17DC01DA00015429624C6FEF68F7_1882747497486589887

PaulKlauser:
Closes #448

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -27,9 +27,7 @@
   <fields>;
 }
 
--keepnames class androidx.navigation.fragment.NavHostFragment
--keepnames class com.willowtree.vocable.room.Category
--keepnames class com.willowtree.vocable.room.Phrase
+-keepnames class * extends android.os.Parcelable
 
 -dontwarn com.google.ar.sceneform.animation.AnimationEngine
 -dontwarn com.google.ar.sceneform.animation.AnimationLibraryLoader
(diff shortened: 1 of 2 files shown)


app/proguard-rules.pro
```diff
@@ -27,9 +27,7 @@
   <fields>;
 }
 
--keepnames class androidx.navigation.fragment.NavHostFragment
--keepnames class com.willowtree.vocable.room.Category
--keepnames class com.willowtree.vocable.room.Phrase
+-keepnames class * extends android.os.Parcelable
```
mattttvaughn, line 30: could this retain anything parcelable even in dependencies and stuff? I'm not an app size hound but we should check that apk size doesn't go up 10MB or anything after this

mattttvaughn (approved):

app/proguard-rules.pro
PaulKlauser: The nice thing about `keepnames` is that all it does is preserve names of classes that **weren't** removed during shrinking. It doesn't keep any classes that otherwise would have been removed during shrinking.

See: https://www.guardsquare.com/manual/configuration/usage#keepoptions

PR #478, opened 2024-01-17
Implement deleting preset/saved categories

PaulKlauser:
Note: this is a separate concept from "hidden"

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/room/PresetCategoryDeleted.kt
+++ b/app/src/main/java/com/willowtree/vocable/room/PresetCategoryDeleted.kt
@@ -1,0 +1,8 @@
+package com.willowtree.vocable.room
+
+import androidx.room.ColumnInfo
+
+data class PresetCategoryDeleted(
+    @ColumnInfo(name = "category_id") val categoryId: String,
+    @ColumnInfo(name = "deleted") val deleted: Boolean
+)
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/willowtree/vocable/room/PresetCategoryDeleted.kt
```diff
@@ -0,0 +1,8 @@
+package com.willowtree.vocable.room
+
+import androidx.room.ColumnInfo
+
+data class PresetCategoryDeleted(
+    @ColumnInfo(name = "category_id") val categoryId: String,
+    @ColumnInfo(name = "deleted") val deleted: Boolean
+)
```
mattttvaughn, line 8: I had not realized you could interface with Room like this 🤯 I would've just written another SQL statement

mattttvaughn: this is so good!!!

mattttvaughn (approved):

app/src/main/java/com/willowtree/vocable/room/PresetCategoryDeleted.kt
PaulKlauser: Yea this is pretty poorly document, but there's some info here: https://developer.android.com/reference/androidx/room/Update !

Issue #467, opened 2024-01-10
[SPIKE] Determine source of truth for translations


PaulKlauser:
We use Crowdin for translations, and currently a number of our string resources for other languages just feature English.

Going to Crowdin, we have multiple projects, Android, iOS, a common project. It's unclear what source of truth we should be using for pulling translations into the app. We need this before going live.

Outcome:
* We know and document which Crowdin project to pull from, and we are confident in its content quality. 
* We remove projects that are not relevant or accurate.

kaffn8ed27:
> We know and document which Crowdin project to pull from, and we are confident in its content quality.
* Android should use the Android project to translate in-app strings. Some strings differ across platforms, hence the need for separate Android and iOS projects.
* Quality is, as far as I can tell, going to be up to whomever translates the strings for us (this being a crowdsourced translation service)
> We remove projects that are not relevant or accurate.
* All of the existing projects are relevant

kaffn8ed27:
@PaulKlauser in what forum should I document the above for posterity?

PR #480, opened 2024-01-18
Separate Tables for Preset/Stored Phrases

mattttvaughn:
Creates a new table for PresetPhraseDto, and creates classes which will be responsible for preset/stored phrase editing/consumption. This will allow us to use a schema for preset phrases with resource IDs, avoiding workarounds that add tech debt to the app.

None of the changes here should affect the existing functionality in the app. The only exception is changes to the DB schema- we have agreed to hold off on writing migrations until we better understand what our schema will be at next release- to prevent ourselves from wasting time writing temporary workarounds.

Part of the work for #467. Following this MR we will:
 * Populate the PresetsDB using resource IDs to have translations that _just work_

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/room/PresetPhrasesRepository.kt
+++ b/app/src/main/java/com/willowtree/vocable/room/PresetPhrasesRepository.kt
@@ -1,0 +1,7 @@
+package com.willowtree.vocable.room
+
+// TODO: MPV #467- this will take over the responsibilities for storing and fetching preset
+//                 phrases from the [ILegacyCategoriesAndPhrasesRepository]
+interface PresetPhrasesRepository {
+    suspend fun addPhrase(phrase: PresetPhraseDto)
+}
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/willowtree/vocable/room/PresetPhrasesRepository.kt
```diff
@@ -0,0 +1,7 @@
+package com.willowtree.vocable.room
+
+// TODO: MPV #467- this will take over the responsibilities for storing and fetching preset
+//                 phrases from the [ILegacyCategoriesAndPhrasesRepository]
+interface PresetPhrasesRepository {
+    suspend fun addPhrase(phrase: PresetPhraseDto)
```
PaulKlauser, line 6: This might not even need to be part of the public interface, seeing as users can't actually _add_ a preset phrase. We can merge it now, I'm not concerned, but wanted to point that out.

PaulKlauser (approved):
Love it. Great small change to start this work 👏 🐋

app/src/main/java/com/willowtree/vocable/room/PresetPhrasesRepository.kt
mattttvaughn: Yea removed already in my current work in progress 😅  just felt weird having an empty interface

PR #484, opened 2024-01-31
Handle Updating Last Spoken Time Across Preset/Stored Tables

mattttvaughn:
Doing the smallest possible change to PhrasesUseCase as we start to transition to using the separated preset/stored phrase database tables. This sets up infrastructure to test PhrasesUseCase with backing by an in-memory room database as well.


app/src/androidTest/java/com/willowtree/vocable/utility/StubLegacyCategoriesAndPhrasesRepository.kt
PaulKlauser, line 1: Nice guard against accidentally calling the old stuff. I hadn't thought of this, cool!

PaulKlauser (changes requested):
As discussed in Slack, let's remove the lastSpokenDate field from the Phrase models beyond the DTO. This may block you from being able to test the update right now until you implement getPhrases(), but we'll come back to it.

PR #485, opened 2024-01-31
Crowdin/remove translations sync

kaffn8ed27:
Description: 
Removes all translated strings. They should be restored to main when we do a CrowdIn sync.


PaulKlauser (approved):
🤞

PR #486, opened 2024-01-31
Adds new translated files from CrowdIn (manually)

kaffn8ed27:
Description: 
Adds translations downloaded from CrowdIn (from nonexistent "develop" branch) so that CrowdIn can point them to main branch


app/src/main/res/values-zh-rTW/strings.xml
```diff
@@ -0,0 +1,151 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <string name="app_name">Vocable</string>
+  <string name="select_something">Hanxinl75@gmail. com</string>
+  <string name="phrases_page_number">hanxinl75@gmail.com</string>
```
mattttvaughn, line 5: I know this just came from Crowdin but this doesn't seem right lol

mattttvaughn:
Looks good to me. Happy to approve as-is, it's a 10000% percent improvement to how it was previously. We can get that thing I commented on fixed on the Crowdin side or here, whatever makes sense from your end

mattttvaughn (approved):

app/src/main/res/values-zh-rTW/strings.xml
kaffn8ed27: Oh that _is_ weird, I hadn't actually noticed it before...I'll make sure it gets deleted. Actually, that'll make for a nice test of the CrowdIn integration

Issue #419, opened 2023-09-14
Stop using View.tag on CategoryButton


PaulKlauser:
View.tag is typed as an Object, which is problematic since you get into type checking and a lack of guarantees that the field's been set correctly.

Let's give CategoryButton a property of `Category` that's more explicit, and the button can also then be responsible for setting up its action and display based on that Category.

PR #494, opened 2024-02-06
419: Stop using View.tag on CategoryButton

kaffn8ed27:
Description: 
Adds category field to CategoryButton & refactors button action/display logic around new field

Screenshots: 
n/a, if UI changes I've done something wrong

- [x] Regression Testing

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/presets/CategoriesFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/presets/CategoriesFragment.kt
@@ -8,7 +8,6 @@ import android.widget.LinearLayout
 import androidx.core.view.children
 import androidx.core.view.isInvisible
 import androidx.core.view.isVisible
-import androidx.lifecycle.Observer
 import com.willowtree.vocable.BaseFragment
 import com.willowtree.vocable.BindingInflater
 import com.willowtree.vocable.R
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/willowtree/vocable/presets/CategoriesFragment.kt
```diff
@@ -58,19 +55,16 @@ class CategoriesFragment : BaseFragment<CategoriesFragmentBinding>() {
...
             with(categoryButton.root) {
-                tag = category
-                action = {
-                    viewModel.onCategorySelected(category.categoryId)
-
-                }
-                text = localizedResourceUtility.getTextFromCategory(category)
+                setUpDisplayAndAction(
+                    category = category
+                ) { viewModel::onCategorySelected }
```
mattttvaughn, line 60: I'm not 100% certain that `viewModel::onCategorySelected` gets called when the lambda gets invoked here- double-check for me?

I think `{ viewMode.onCategorySelected() }` would do it for sure, uncertain with the `::` syntax

IanCrossCD: I think your instinct is correct. It would need to be passed as a parameter, not in a lambda body.

kaffn8ed27: Looks like it's getting called whenever I switch categories. I think that when CategoryButton sets up its action with this funciton passed in (line 40), it makes the actual call. Behavior (user-facing and with breakpoints) seems to match main, which is the goal

kaffn8ed27: Hmmm, I'm possibly not looking at the right thing here, since tapping the button doesn't seem to work. I'll revert to passing it the other way - unfortunately this means even less gets refactored into the button itself

mattttvaughn (approved):

kaffn8ed27:
@PaulKlauser I'm not 💯 sure if this was quite everything you had in mind for #419. I'm happy to get together and see how much further we can take this, if my hunch is correct.

PR #495, opened 2024-02-06
Allow Updating of Preset Phrases

mattttvaughn:
Description: 
Allows preset phrases to be edited. When a preset phrase is edited, it is marked as hidden in the preset phrases table and is "shadowed" (to use Kotlin terminology) by the phrase in the custom phrases table. 

Screenshots: 
[vocable-edit-presets.webm](https://github.com/willowtreeapps/vocable-android/assets/58408581/75044fbd-44f8-4f97-91e7-acb3654896ef)

### DIFF
--- a/app/src/androidTest/java/com/willowtree/vocable/PhrasesUseCaseTest.kt
+++ b/app/src/androidTest/java/com/willowtree/vocable/PhrasesUseCaseTest.kt
@@ -156,8 +156,7 @@ class PhrasesUseCaseTest {
                     phraseId = "1",
                     localizedUtterance = testLocalesWithText,
                     sortOrder = 0,
-                    lastSpokenDate = null,
-                    parentCategoryId = "category"
+                    lastSpokenDate = null
                 )
             ),
             useCase.getPhrasesForCategory("category")
--- a/app/src/androidTest/java/com/willowtree/vocable/PhrasesUseCaseTest.kt
+++ b/app/src/androidTest/java/com/willowtree/vocable/PhrasesUseCaseTest.kt
@@ -12,6 +12,7 @@ import com.willowtree.vocable.room.RoomStoredPhrasesRepository
 import com.willowtree.vocable.room.VocableDatabase
 import com.willowtree.vocable.utility.FakeDateProvider
 import com.willowtree.vocable.utility.StubLegacyCategoriesAndPhrasesRepository
+import com.willowtree.vocable.utils.UUIDProvider
 import com.willowtree.vocable.utils.locale.LocalesWithText
 import junit.framework.Assert.assertEquals
 import kotlinx.coroutines.test.runTest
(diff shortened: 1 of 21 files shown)


app/src/androidTest/java/com/willowtree/vocable/PhrasesUseCaseTest.kt
```diff
@@ -152,9 +157,47 @@ class PhrasesUseCaseTest {
...
+    fun updatePhrase_updatesPresetPhrase() = runTest {
+        val useCase = createUseCase()
+        presetPhrasesRepository.populateDatabase()
+        val phraseId = "category_123_0"
+        val localizedUtterance = LocalesWithText(mapOf("en" to "updated text"))
+
+        useCase.updatePhrase(phraseId, localizedUtterance)
+
+        val newCustomPhrase = useCase.getPhrasesForCategory(PresetCategories.USER_KEYPAD.id)
+            .singleOrNull { it.phraseId == phraseId } as CustomPhrase
```
PaulKlauser, line 200: Nice! Is this `single` how you're asserting that the preset phrase is no longer being returned?

app/src/main/java/com/willowtree/vocable/PhrasesUseCase.kt
```diff
@@ -35,12 +39,40 @@ class PhrasesUseCase(
...
+                        creationDate = dateProvider.currentTimeMillis(),
+                        lastSpokenDate = phrase.lastSpokenDate,
+                        localizedUtterance = localizedUtterance,
+                        sortOrder = phrase.sortOrder
+                    )
+                )
+
+            }
+            null -> throw IllegalArgumentException("Phrase with id $phraseId not found")
+        }
```
PaulKlauser, line 70: I like

app/src/main/java/com/willowtree/vocable/presets/Phrase.kt
```diff
@@ -12,6 +12,7 @@ sealed interface Phrase : Parcelable {
     val phraseId: String
     val sortOrder: Int
     val lastSpokenDate: Long?
+    val parentCategoryId: String?
```
PaulKlauser, line 15: Just to make sure I understand, we needed this so that we can create the custom Phrase with the same category as the edited preset? Did this field need to be added to Custom and Preset?

mattttvaughn: ah yea true. I was originally updating the custom phrase via a repo.addPhrase() and needed all of the properties but now that it's using the cool `@Update` tech we only need the properties being changed. removed!

```
storedPhrasesRepository.updatePhraseLocalizedUtterance(
    phraseId = phraseId,
    localizedUtterance = localizedUtterance,
)
```

app/src/androidTest/java/com/willowtree/vocable/PhrasesUseCaseTest.kt
mattttvaughn: Correct- It was more like the `single` was me assuming/asserting there was only one phrase being returned, and that the one being returned has a localizedUtterance (was not a preset)

Changed it to be a little bit more explicit- using an assert to make it easier to track down test failures on this

PaulKlauser: Nice, I like the explicit assertion, I think that does make it more clear!

PaulKlauser (approved):

PR #499, opened 2024-02-12
Chore/452 abstract over saved preset categories 12

PaulKlauser:
Resource IDs are not stable. We were also needing to differentiate the type of category to get the category's name, this way that's abstracted away. 

We still need another field to differentiate from the UI's idea of "hidden" and the idea of a preset category being permanently hidden. Until then, modifying the name of a preset category will result in it still appearing in settings.

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/presets/PresetCategories.kt
+++ b/app/src/main/java/com/willowtree/vocable/presets/PresetCategories.kt
@@ -3,11 +3,11 @@ package com.willowtree.vocable.presets
 import com.willowtree.vocable.R
 
 enum class PresetCategories(val id: String, val initialSortOrder: Int) {
-    GENERAL("preset_general_category_id", 0),
-    BASIC_NEEDS("preset_basic_needs_category_id", 1),
-    PERSONAL_CARE("preset_personal_care_category_id", 2),
-    CONVERSATION("preset_conversation_category_id", 3),
-    ENVIRONMENT("preset_environment_category_id", 4 ),
+    GENERAL("preset_general", 0),
+    BASIC_NEEDS("preset_basic_needs", 1),
+    PERSONAL_CARE("preset_personal_care", 2),
+    CONVERSATION("preset_conversation", 3),
+    ENVIRONMENT("preset_environment", 4 ),
     USER_KEYPAD("preset_user_keypad", 5),
     RECENTS("preset_recents", 6),
     @Deprecated("This is being filtered out from the UI already. Remove this.")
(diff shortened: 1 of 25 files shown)


app/src/main/java/com/willowtree/vocable/room/Version7Migration.kt
```diff
@@ -0,0 +1,10 @@
+package com.willowtree.vocable.room
+
+import androidx.room.DeleteColumn
+import androidx.room.migration.AutoMigrationSpec
+
+@DeleteColumn(
+    tableName = "Category",
+    columnName = "resource_id"
+)
+class Version7Migration : AutoMigrationSpec
```
mattttvaughn, line 10: ooh more new Room tech

app/src/main/java/com/willowtree/vocable/presets/PresetCategories.kt
```diff
@@ -3,11 +3,11 @@ package com.willowtree.vocable.presets
...
-    GENERAL("preset_general_category_id", 0),
-    BASIC_NEEDS("preset_basic_needs_category_id", 1),
-    PERSONAL_CARE("preset_personal_care_category_id", 2),
-    CONVERSATION("preset_conversation_category_id", 3),
-    ENVIRONMENT("preset_environment_category_id", 4 ),
+    GENERAL("preset_general", 0),
+    BASIC_NEEDS("preset_basic_needs", 1),
+    PERSONAL_CARE("preset_personal_care", 2),
+    CONVERSATION("preset_conversation", 3),
+    ENVIRONMENT("preset_environment", 4 ),
```
mattttvaughn, line 10: nice to see this simplifying the getNameId() stuff at the same time 😁

app/src/test/java/com/willowtree/vocable/FakeCategoriesUseCase.kt
```diff
@@ -72,7 +88,11 @@ class FakeCategoriesUseCase : ICategoriesUseCase {
         _categories.update { categories ->
             categories.map {
                 if (it.categoryId == categoryId) {
-                    it.copy(hidden = hidden)
+                    when(it) {
+                        is Category.StoredCategory -> it.copy(hidden = hidden)
+                        is Category.PresetCategory -> it.copy(hidden = hidden)
+                        is Category.Recents -> it.copy(hidden = hidden)
```
mattttvaughn, line 94: ah I was thinking the few instances that looks like this were a little smelly but looks like this is not a problem in the real impl because of  `withHidden(...)` and friends. 

fine to keep this as-is imo since we're deprecating this bad boy

mattttvaughn (approved):

app/src/test/java/com/willowtree/vocable/FakeCategoriesUseCase.kt
PaulKlauser: Ah shoot, I forgot about the `withHidden()`, you just reminded me why I created that to begin with!

Agreed, let's keep this as-is since it's going to get deleted anyway, and this approach isn't really harming anyone. Agreed though, it is a little smelly.

PR #506, opened 2024-02-15
Allow Deletion of Preset Phrases

### LINKED ISSUE #505: Delete preset and saved categories
As a user, I want to be able to delete a category, whether it's preset, or saved.

## Scenario 1 - Preset
Given: I am editing a preset category
When: I delete that category
Then: The category and its associated phrases no longer show in settings or the main screen of the app

## Scenario 2 - Custom
Given: I am editing a custom category
When: I delete that category
Then: The category and its associated phrases no longer show in settings or the main screen of the app

### Note
This depends on #504

mattttvaughn:
Allows Preset phrases to be deleted from the app. Custom phrases were already deletable, so the UI is the same, but now preset phrases can be deleted too.

[delete_presets.webm](https://github.com/willowtreeapps/vocable-android/assets/58408581/b2b25aee-a438-41b7-9465-3aa6839437a8)

Closes #505

### DIFF
--- a/app/src/androidTest/java/com/willowtree/vocable/PhrasesUseCaseTest.kt
+++ b/app/src/androidTest/java/com/willowtree/vocable/PhrasesUseCaseTest.kt
@@ -202,4 +202,39 @@ class PhrasesUseCaseTest {
             .singleOrNull { it.phraseId == phraseId } as CustomPhrase
         assertEquals(localizedUtterance, newCustomPhrase.localizedUtterance)
     }
+
+    @Test
+    fun deletePhrase_deletesFromStoredRepository() = runTest {
+        val useCase = createUseCase()
+        val customPhraseId = "1"
+        storedPhrasesRepository.addPhrase(
+            PhraseDto(
+                phraseId = customPhraseId,
+                localizedUtterance = testLocalesWithText,
+                parentCategoryId = PresetCategories.GENERAL.id,
+                creationDate = 0L,
+                lastSpokenDate = 100L,
+                sortOrder = 0
+            )
+        )
+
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/willowtree/vocable/presets/Phrase.kt
```diff
@@ -34,6 +34,7 @@ data class PresetPhrase(
     override val phraseId: String,
     override val sortOrder: Int,
     override val lastSpokenDate: Long?,
+    val deleted: Boolean,
```
PaulKlauser, line 37: Thoughts on filtering at the repository level, and then not needing the concept of `deleted` being in any of our domain models?

app/src/androidTest/java/com/willowtree/vocable/PhrasesUseCaseTest.kt
```diff
@@ -202,4 +202,39 @@ class PhrasesUseCaseTest {
...
+        val useCase = createUseCase()
+        val presetPhraseId = "category_123_0"
+        presetPhrasesRepository.populateDatabase()
+
+        useCase.deletePhrase(presetPhraseId)
+
+        val presetPhrase = useCase.getPhrasesForCategory(PresetCategories.USER_KEYPAD.id)
+            .firstOrNull { it.phraseId == presetPhraseId }
+        assertEquals(null, presetPhrase)
+    }
```
PaulKlauser, line 239: 💯

PaulKlauser (approved):
Merging so we can continue on with categories, but I would like to close the loop on the `deleted` field being present in the Phrase model

PR #509, opened 2024-02-20
Handle sorting in CategoriesUseCase

PaulKlauser:
We don't need to actually update sort order when adding a category, if we're cool with it always being at the end of the list. Additionally, we can check for hidden and do that sorting at the same layer, further simplifying the logic.

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
+++ b/app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
@@ -10,6 +10,8 @@ import com.willowtree.vocable.utils.locale.LocaleProvider
 import com.willowtree.vocable.utils.locale.LocalesWithText
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.map
 
 class CategoriesUseCase(
     private val uuidProvider: UUIDProvider,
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
```diff
@@ -75,13 +85,15 @@ class CategoriesUseCase(
...
         storedCategoriesRepository.upsertCategory(
             Category.StoredCategory(
                 uuidProvider.randomUUIDString(),
                 LocalesWithText(mapOf(Pair(localeProvider.getDefaultLocaleString(), categoryName))),
                 false,
-                sortOrder
+                allCategories.size
             )
         )
     }
```
mattttvaughn, line 99: theoretically if we have 7 presets

```
sortOrders: [0, 1, 2, 3, 4, 5, 6]
```

add one category

```
sortOrders: [0, 1, 2, 3, 4, 5, 6, 7]
```

then delete a preset:
```
sortOrders: [1, 2, 3, 4, 5, 6, 7]
```

if we wanted to add another category then we end up with two categories with the same sortOrder:

```
sortOrders: [1, 2, 3, 4, 5, 6, 7, 7]
```

Seems like it could cause (very minor) problems/non-determinism. I don't forsee any big problems but maybe we could avoid the problem altogether by doing something like `categories.maxBy { it.sortOrder } + 1` instead?

PaulKlauser: Yea that's a really good catch. Trying to thread the needle of not needing logic that maintains perfectly continuous numbers, but also not breaking things

PaulKlauser: There are also definitely issues with re-sorting in the app right now. I'm going to merge and continue moving forward so that this logic is at least out of the add category ViewModel, and follow up to fix things in their new home.

PR #510, opened 2024-02-26
Implement moving category sort order up and down

PaulKlauser:


app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
```diff
@@ -21,6 +21,17 @@ class CategoriesUseCase(
...
+        storedCategoriesRepository.getAllCategories(),
+        presetCategoriesRepository.getPresetCategories()
+    ) { storedCategories, presetCategories ->
+        storedCategories.map { it.asCategory() } + presetCategories
+    }.map {
+        it.sortedBy { category ->
+            category.sortOrder
+        }
+    }
+
```
mattttvaughn, line 34: this doesn't look right- old code? the variable is unused and the behavior (sorted) doesn't match the name (unsorted)

app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
```diff
@@ -106,4 +117,46 @@ class CategoriesUseCase(
...
+                    categoryId = category.categoryId,
+                    sortOrder = nextCategory.sortOrder
+                ),
+                CategorySortOrder(
+                    categoryId = nextCategory.categoryId,
+                    sortOrder = category.sortOrder
+                )
+            )
+        )
+    }
```
mattttvaughn, line 160: I dig the swapping sortOrders approach- seems robust 👍

app/src/main/java/com/willowtree/vocable/CategoriesUseCase.kt
PaulKlauser: This was an experiment that leaked through, good catch!

PR #515, opened 2024-03-18
Add notion of category pages

PaulKlauser:
The idea is that this data can be observed by the page fragments, so that the pager adapter's content never has to actually change unless the page count changes. This should help us mitigate the display issues that are caused by the adapter rebinding any time the category list changes.

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/settings/EditCategoriesViewModel.kt
+++ b/app/src/main/java/com/willowtree/vocable/settings/EditCategoriesViewModel.kt
@@ -9,9 +9,13 @@ import com.willowtree.vocable.IPhrasesUseCase
 import com.willowtree.vocable.presets.Category
 import com.willowtree.vocable.presets.Phrase
 import com.willowtree.vocable.room.CategorySortOrder
+import com.willowtree.vocable.settings.editcategories.EditCategoriesPage
 import com.willowtree.vocable.utils.ILocalizedResourceUtility
 import kotlinx.coroutines.flow.first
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/willowtree/vocable/settings/EditCategoriesViewModel.kt
```diff
@@ -25,6 +29,17 @@ class EditCategoriesViewModel(
...
+        val pageSize = 8
+        val pageCount = ceil(categories.size / pageSize.toFloat()).toInt()
+
+        (0 until pageCount).map {
+            val start = it * pageSize
+            val end = min((it + 1) * pageSize, categories.size)
+            EditCategoriesPage(categories.subList(start, end))
+        }
+    }
+
```
mattttvaughn, line 42: seems like this could be a good case for  using `List.chunked(n)` 👀 

```
val categoryPages = categoriesUseCase.categories().map { categories ->
    val pageSize = 8
    categories.chunked(pageSize) { 
        EditCategoriesPage(it)
    }
}
```

mattttvaughn (approved):
lgtm. one potential improvement to your discretion

app/src/main/java/com/willowtree/vocable/settings/EditCategoriesViewModel.kt
PaulKlauser: I knew there was a better API for this! I couldn't remember what it was. You've saved the day, will definitely implement that

Issue #416, opened 2023-09-06
Remove duplicate category state in EditCategoriesViewModel


PaulKlauser:
`EditCategoriesViewModel` currently holds separate state for reordering as well as adding and removing categories. This was necessitated by the fact that the ViewPager/List that displays categories gets reset every time the list changes, so reordering categories would change pages and be a jarring experience for the user.

However, this workaround creates misdirection as there are now multiple sources of truth for the list of categories. There should be a way to make the ViewPager/List maintain its position even through reordering. `ListAdapter`, and its associated diffing logic is likely the solution, and this might even enable reordering animations, which would further improve UX.

On completion of this ticket, `EditCategoriesViewModel`, should have a single source of truth for the categories list within.

PR #524, opened 2024-04-11
Issue 416: Refactor pager code of edit categories screen

josmith42:
Description: 
* Partial solution for https://github.com/willowtreeapps/vocable-android/issues/416
* `VocableFragementStateAdapter`:
  * Fix return value of `getItemCount()`
  * Replace `notifyDataSetChanged()` with `notifyItemInserted()` and `notifyItemRemoved()`
* `EditCategoriesFragment`:
  * Use `post {}` for setting `categoryPageNumber` (wasn't always getting set correctly before)
  * Move setting of adapter so that it's called once instead of every time data changes
  * Remove broken setting of current page to "middle" (now defaults to first page)
* Make `LegacyFragmentStateAdapter` for use in other subclasses that haven't been changed yet
* `EditCategoriesViewModel`: replace `liveOrderCategoryList` and `liveAddRemoveCategoryList` in favor of a single `categoryList` that uses `categories()` flow from `categoriesUseCase`
* `EditCategoriesListFragment`: fix crash when removing an element caused pager to go down a page

Screenshots: 
n/a

- [ ] Regression Testing

### DIFF
--- a/app/src/test/java/com/willowtree/vocable/FakeCategoriesUseCase.kt
+++ b/app/src/test/java/com/willowtree/vocable/FakeCategoriesUseCase.kt
@@ -5,6 +5,7 @@ import com.willowtree.vocable.room.CategorySortOrder
 import com.willowtree.vocable.utils.locale.LocalesWithText
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.update
 
 @Deprecated("This fake is too complex, tests using it should migrate to integration tests with" +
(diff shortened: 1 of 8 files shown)


mattttvaughn:
Seeing something funky going on here. If you go to the categories page and keep trying to move the bottom most category on a page down, the moving doesn't happy as expected and if you do it enough it'll crash

[vocable_crash.webm](https://github.com/willowtreeapps/vocable-android/assets/58408581/0f13241b-1c2e-4da8-af16-e67126980b17)

app/src/test/java/com/willowtree/vocable/FakeCategoriesUseCase.kt
```diff
@@ -23,7 +24,15 @@ class FakeCategoriesUseCase : ICategoriesUseCase {
     )
 
     override fun categories(): Flow<List<Category>> {
-        return _categories
+        return _categories.map {
```
josmith42, line 27: I had to make this change to get the fake to match the behavior of the real use case. (I know, I know. The real solution is to refactor the test to use the real use case. I am just not prepared to do that as a part of this PR.)

josmith42:
@mattttvaughn I fixed the crash you raised up. Please take a look! 🙏

mattttvaughn (approved):

PR #529, opened 2024-05-14
380: Decreases individual category padding

cinadia:
Description: 
Decrease individual category padding from 10dp to 5dp to prevent cutoff in landscape orientation.

Screenshots: 
image

- [ ] Regression Testing

### DIFF
--- a/app/src/main/res/values/dimens.xml
+++ b/app/src/main/res/values/dimens.xml
@@ -238,7 +238,7 @@
     <dimen name="edit_category_pager_margin_bottom">44dp</dimen>
     <dimen name="edit_individual_category_padding_start">20dp</dimen>
     <dimen name="edit_individual_category_padding_end">0dp</dimen>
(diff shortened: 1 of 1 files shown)


PaulKlauser:
Thanks for the contribution @cinadia ! 😁  This looks good to me as far as fixing the original bug, let's merge it in! Before marking the ticket as complete though, can you look into why the rows are being constrained to the smaller height that resulted in clipping the text in the first place? My expectation would be that the original padding value would only cause an issue if the rows themselves were being forced to be smaller than they need to be, and I think the UX would be better if they were allowed to expand a little.

PaulKlauser (approved):

cinadia:
@PaulKlauser I believe it's because the heights of the individual edit category button is set to the same height as the arrow button, which just isn't enough space given the original padding.

In `land/category_edit_button.xml`: 
image

PR #530, opened 2024-05-15
445: Stores spoken keyboard prompts to Recents

### LINKED ISSUE #445: Custom phrases are not added to Recents list
**Describe the bug**
App allow the creation of custom phrases entered with keyboard. These phrases are not added to the recents list. There is ability to create a custom phrase in each category. Not sure how much utility it adds to have exclusively temporary phrases.

**To Reproduce**
Tap keyboard button on any category except 123 and Recents.
Enter a word to say
Tap Voice button
Tap chat button to return to main screen
Nav to Recents

**Expected behavior**
Any phrase used in the app will be listed in the Recents

**Actual Behavior**
Keyboard entered phases are discarded

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: samsung galaxy s21
 - OS: Android 13

cinadia:
Description: 
Stores keyboard prompts that the user Speaks to the Recents category
Resolves #445 

- [ ] Regression Testing

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/keyboard/KeyboardFragment.kt
+++ b/app/src/main/java/com/willowtree/vocable/keyboard/KeyboardFragment.kt
@@ -5,6 +5,7 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.core.view.children
 import androidx.core.view.isVisible
+import androidx.fragment.app.viewModels
 import androidx.lifecycle.Observer
 import androidx.navigation.fragment.findNavController
 import androidx.recyclerview.widget.GridLayoutManager
@@ -14,11 +15,14 @@ import com.willowtree.vocable.R
 import com.willowtree.vocable.customviews.PointerListener
 import com.willowtree.vocable.databinding.FragmentKeyboardBinding
 import com.willowtree.vocable.keyboard.adapter.KeyboardAdapter
+import com.willowtree.vocable.presets.PresetCategories
 import com.willowtree.vocable.utils.ItemOffsetDecoration
 import com.willowtree.vocable.utils.VocableTextToSpeech
 import java.util.Locale
 
 class KeyboardFragment : BaseFragment<FragmentKeyboardBinding>() {
+
+    private val viewModel: KeyboardFragmentViewModel by viewModels()
 
     override val bindingInflater: BindingInflater<FragmentKeyboardBinding> =
         FragmentKeyboardBinding::inflate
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/willowtree/vocable/keyboard/KeyboardFragment.kt
```diff
@@ -104,10 +108,12 @@ class KeyboardFragment : BaseFragment<FragmentKeyboardBinding>() {
 
         binding.keyboardSpeakButton.action = {
             if (!isDefaultTextVisible()) {
+                val text = binding.keyboardInput.text?.toString() ?: ""
                 VocableTextToSpeech.speak(
                     Locale.getDefault(),
-                    binding.keyboardInput.text?.toString() ?: ""
+                    text
                 )
+                viewModel.addNewPhrase(text, PresetCategories.RECENTS.id)
```
PaulKlauser, line 116: You exposed a flaw in how our categories are set up, nice! We don't actually want to be storing phrases under the "RECENTS" category, instead we determine RECENTS by just looking at the most recently used phrases [PhrasesUseCase#getPhrasesForCategory()](https://github.com/willowtreeapps/vocable-android/blob/492d1409be508a3c6ecd81181985da3d0a75879a/app/src/main/java/com/willowtree/vocable/PhrasesUseCase.kt#L23).

Totally understand how you got here, and we should clean up the API to not allow creating a phrase with "Recent" as the category, could you create a ticket on the board to track that? Up to you whether or not you tackle that, but we should get it written down.

As far as the defect at hand, I don't think we actually want just anything spoken from the keyboard to be saved, and this is corroborated by how the iOS app works. Instead, Android is currently missing functionality where a user can save a phrase directly from the keyboard. Download the Vocable iOS app to your Mac and poke around to see what I mean. Once you feel comfortable, I think we should discard the current bug as "won't-fix" and write up an enhancement to match iOS's functionality here.

PaulKlauser (changes requested):

app/src/main/java/com/willowtree/vocable/keyboard/KeyboardFragment.kt
cinadia: Updated #445 to reflect adding the My Sayings enhancement.

PaulKlauser: @cinadia Agreed!

PR #579, opened 2024-10-02
Update Kotlin and Room to be able to build on AS Ladybug

PaulKlauser:
Ladybug bundles JDK 21, which requires a Kotlin version bump, which in-turn requires a Room bump. No functional changes.


nicholas-cook (approved):

SashankPatel:
Closing this out as the some of these kotlin libraries were updated recently.

PR #583, opened 2025-01-27
Adds feature flags with an initial listen mode flag.

andrewroycarter:
Description: 
* Adds Feature Flags to enable Trunk Based Development. Runtime changing via a debug menu is not present at the moment, but it's intended to be supported by this work if needed.
* Adds initial unused listen mode feature flag
* Adds tests around new Feature Flag functionality.

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/utils/features/FeatureFlags.kt
+++ b/app/src/main/java/com/willowtree/vocable/utils/features/FeatureFlags.kt
@@ -1,0 +1,44 @@
+package com.willowtree.vocable.utils.features
+
+import com.willowtree.vocable.BuildConfig
+import com.willowtree.vocable.utils.IVocableSharedPreferences
+
+/**
+ * Central location for managing feature flags in the application.
+ * Supports both build-time and runtime feature flags.
+ */
+object FeatureFlags {
+    private const val KEY_LISTEN_MODE = "listen_mode"
+
+    /**
+     * Interface for all feature flag implementations.
+     */
+    interface Flag {
+        /**
+         * Whether the feature is currently enabled.
+         */
+        val enabled: Boolean
+    }
+
+    /**
+     * Build-time flags that are controlled via build configuration.
+     */
+    object BuildTimeFlag : Flag {
+        override val enabled: Boolean = BuildConfig.ENABLE_LISTEN_MODE
+    }
+
+    /**
+     * Runtime flags that can be toggled via SharedPreferences.
+     */
+    class RuntimeFlag(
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/willowtree/vocable/utils/features/FeatureFlags.kt
```diff
@@ -0,0 +1,44 @@
...
+        override val enabled: Boolean = BuildConfig.ENABLE_LISTEN_MODE
+    }
+
+    /**
+     * Runtime flags that can be toggled via SharedPreferences.
+     */
+    class RuntimeFlag(
+        private val sharedPreferences: IVocableSharedPreferences,
+        private val key: String,
+        private val defaultValue: Boolean
```
JonathanNye, line 36: The patterns for _defining_ flags and _accessing_ their state feels a little off here.

Imagine I'm writing code in some biz logic somewhere. If I want to check a runtime flag, it seems like I'd do something like this (I just borrowed it from your test):

`FeatureFlags.RuntimeFlag(sharedPrefs, "test_flag", true)`

The onus is on me, random hunk of biz logic, to know the specific key for the feature and what its default should be. It seems like both of those ideas could be baked into some kind of sealed/enum-y type that defines individual flags and _that_ could be passed to a thing that determines the state for you.

Something like...
```
enum class RuntimeFlag(key: String, default:Boolean) {
    LISTEN_MODE("listen_mode", false)
}
```
and then break the thing that _resolves_ the flag out...
```
class RuntimeFlags(/* yadda, Shared Prefs probably */) {
    operator fun get(flag: FeatureFlag): Boolean = /* yadda, read from disk */
}
```
and in situ... (I got cute with the indexing syntax operator, that's optional)
```
val runtimeFlags = RuntimeFlags(/* yadda */)
val isMyFeatureEnabled = runtimeFlags[FeatureFlag.LISTEN_MODE]
```

app/src/main/java/com/willowtree/vocable/utils/features/FeatureFlags.kt
```diff
@@ -0,0 +1,44 @@
...
+         * Whether the feature is currently enabled.
+         */
+        val enabled: Boolean
+    }
+
+    /**
+     * Build-time flags that are controlled via build configuration.
+     */
+    object BuildTimeFlag : Flag {
+        override val enabled: Boolean = BuildConfig.ENABLE_LISTEN_MODE
```
JonathanNye, line 27: How does this scale when there's more than one build-time flag?

PR #612, opened 2026-07-27
#611: Fix Numbers preset category grid shape and keypad ordering

### LINKED ISSUE #611: [BUG]: Fix "Numbers" preset category -  tablet grid shape for the "123" numeric preset category to match approved Figma designs
BUG: 
**Repo:** vocable-android. A numeric preset category
**[Figma](https://www.figma.com/design/cpKi0vTudc6h9zzRkVVbay/Vocable---Current--Copy-?node-id=64-2146&t=xXzHaX7ojqDGHEHr-4) (see all screens below this one)**

Ticket updated — dropped the label ~~/order~~ items entirely, scoped now to just the tablet grid-shape regression:

---
**Implementation breakdown (1-day segments)**

**Day 1 — Setup, orientation, repro**
- [x] App builds and runs locally from a clean clone.
- [x] Engineer can point to the exact line in `PresetsScreen.kt` where `phraseColumns` branches on `USER_KEYPAD` but `phraseRows` doesn't.
- [x] Bug reproduced and screenshotted on a tablet emulator (landscape + portrait), confirming the incorrect grid shape.

**Day 2 — Implement the fix**
- [x] Tablet landscape renders the "123" category as 3 cols × 4 rows, no blank cells.
- [x] Tablet portrait renders the "123" category as 3 cols × 4 rows, no blank cells.
- [x] Ensure item order (0,1,...,9,Yes,No) **matches** from the Figma frames, with 0 at the bottom
- [x] New/updated unit test passes locally, covering the fixed row count.
- [x] `./gradlew testDebug` passes with no new failures.

**Day 3 — Regression QA, PR, review**
- [ ] No regressions on any other preset category across all 4 breakpoints (screenshots/recording attached to PR).
- [ ] PR opened against `main`, CI green, maintainer review requested.
- [ ] All review feedback addressed and PR approved/merged.

---

### Context
Label ("123") ~~and item order (`0,1,...,9,Yes,No`)~~ stay as-is — explicitly out of scope. This ticket covers only the tablet grid-layout regression.

### Production status
**Unreleased regression**, no current user impact. Introduced by the Jetpack Compose rewrite (PR #605, merged Apr 17, 2026); Play Store is still on the Jan 10, 2024 pre-Compose build. Should be fixed before the rewrite ships.

### Problem
1. Tablet landscape (`64:2146`): Figma shows 3 cols × 4 rows; current code renders 6 cols × 3 rows with a blank row (`values-sw600dp-land/integers.xml`).
2. Tablet portrait (`64:2175`): Figma shows 3 cols × 4 rows; current code renders 3 cols × 5 rows with a blank row (`values-sw600dp/integers.xml`).

Handset landscape (6×2) and handset portrait (3×4) already match — untouched. Root cause: `PresetsScreen.kt` only has a one-liner **column** override for `USER_KEYPAD`; there's no equivalent **row** override, so rows fall back to the generic per-breakpoint value.

### Proposed approach
Add a `phrases_rows_one_liner_phrases` integer override (4 for both tablet breakpoints, matching existing values for phone breakpoints so they're unaffected), and branch `phraseRows` in `PresetsScreen.kt` on `USER_KEYPAD.id` the same way `phraseColumns` already does.
___
         **ENGINEERING NOTES: **
               Beyond the ticket text, here's what I'd hand an engineer directly — things that came out of the investigation but wouldn't naturally live in ticket fields:

**How to reproduce/verify locally**
Android resource qualifiers only kick in on a device/emulator whose `smallestWidth` actually matches — you can't see the tablet-landscape bug on a phone emulator. Use a tablet AVD (e.g. Pixel Tablet, or any `sw600dp`+ profile) in both orientations, navigate to the "123" category, and count cells. `PresetsScreen.kt` already has `@Preview` annotations at the bottom of the file (`PresetsScreenPreview2` specifically uses `USER_KEYPAD`) — worth adding a `sw600dp`-equivalent preview device spec there so this is checkable in Android Studio's preview pane without a physical/emulated tablet at all.

**Exact current (wrong) values, so they don't have to re-derive them**
- `app/src/main/res/values-sw600dp-land/integers.xml`: `phrases_columns_one_liner_phrases=6`, `phrases_rows=3` → renders 6×3.
- `app/src/main/res/values-sw600dp/integers.xml`: `phrases_columns_one_liner_phrases=3`, `phrases_rows=5` → renders 3×5.
- Target for both: 4 rows. The fix is adding a `phrases_rows_one_liner_phrases` resource (doesn't exist yet) and wiring it into `PresetsScreen.kt` around line 120, mirroring the existing `phraseColumns` branch at lines 115-119.

**Evidence trail for the "unreleased regression" claim**, so they can verify without redoing the archaeology:
- Compose rewrite: PR #605, merge commit `c568a1e8` (Apr 17, 2026).
- Last confirmed Play Store build: matches commit `fbac2b72` (Jan 10, 2024) — pre-Compose, still `CategoriesFragment.kt`.
- Only release tag in the repo is `v1.3.0` (2020); `.github/workflows/ps-release.yml` triggers only on pushing a `*.*.*` tag, and none has been pushed since.
- Confirmed live via the Play Store listing itself (`play.google.com/store/apps/details?id=com.willowtree.vocable` → "Updated on Jan 10, 2024").

**Explicitly not a mistake:** label ("123") ~~and item order (`0,1,...,9,Yes,No`)~~ differ from the Figma frames too, but that's an intentional decision from this conversation — don't "fix" those to match Figma.

**Caveats:**
- I couldn't check GitHub's project board for a duplicate issue (connector wasn't authorized in this session) — worth a quick search before starting.
- I verified handset landscape/portrait already match Figma, so there's no need to touch those breakpoints — only `values-sw600dp` and `values-sw600dp-land` need the new row-override value.

Tablet landscape (64:2146) and tablet portrait (64:2175) — both show the correct 3×4 grid to attach as visual reference on the ticket.

Image

Image

___

## Acceptance criteria
- [ ] Tablet landscape/portrait render 3×4 for the "123" category.
- [ ] Handset layouts, pagination, other categories, tap-to-speak/Recents all unchanged.
- [ ] Label ~~and item order~~ remain unchanged.
- [ ] item order (0,1,...,9,Yes,No) **matches** from the Figma frames, with 0 at the bottom
- [ ] New/updated unit test passes locally, covering the fixed row count.
- [ ] `./gradlew testDebug` passes with no new failures.
- [ ]  PR opened against `main`, CI green, maintainer review requested.
- [ ] All review feedback addressed and PR approved/merged.

### Test notes
- Unit: extend `EditCategoriesViewModelTest.kt` / `PhrasesUseCaseTest.kt` for row-count resolution.
- On-device: verify tablet landscape + portrait visually, plus spot-check a non-keypad category for regressions.

### Label
`bug`

### Risks / guardrails
- PR targets `develop`, needs maintainer (Chris) review.
- Not urgent (no live impact) but should land before next Play Store release.
- Couldn't check GitHub board for duplicates — connector still unauthorized.

---

Plan file at [https-www-figma-com-design-cpki0vtudc6h9-binary-wilkinson.md](file:///Users/rhyslutsky/.claude/plans/https-www-figma-com-design-cpki0vtudc6h9-binary-wilkinson.md) reflects this. Still holding — let me know when to file or implement.

kaylaheath-ux:
## Summary

The "123" (Numbers) category rendered as a 6×3 / 3×5 grid with a blank row instead of the 3×4 keypad the designs call for. Fixing the grid shape exposed two further problems behind it: the item order wasn't a keypad either, and the mechanism that seeds preset phrases could never deliver a correction to anyone who had already opened the app.

### Grid shape

- `phraseRows` in `PresetsScreen.kt` had no `USER_KEYPAD` override (unlike `phraseColumns`), so it always fell back to the generic per-breakpoint row count. Added `phrases_rows_one_liner_phrases` to every breakpoint `integers.xml` and branched `phraseRows` on `USER_KEYPAD.id` the same way `phraseColumns` already does.
- Fixed `phrases_columns_one_liner_phrases` in `values-sw600dp-land/integers.xml` (was `6`, needs `3` for the 3×4 target).
- Handset values are unchanged — the new resource values match the generic ones there.
- Added tablet-width `@Preview` specs to `PresetsScreenPreview2` and corrected that preview's phrase list, which had duplicate ids and stale labels.

### Item order

`R.array.category_123` was declared `0 1 2 … 9 Yes No`. The grid fills row-major, so that renders `0 1 2 / 3 4 5 / …`, not a phone keypad. Reordered to `1 … 9 0 No Yes`, which renders `1 2 3 / 4 5 6 / 7 8 9 / 0 No Yes` at the 3-column breakpoints and `1 2 3 4 5 6 / 7 8 9 0 No Yes` at the 6-column ones — both matching the approved designs (Figma 64:2146 / 64:2175).

### Making the arrays authoritative

Reordering an array on its own changes nothing for an existing install, because `sort_order` is only written when a phrase is first seeded:

- `populateDatabase()` was gated behind a `firstTime` shared pref, so it never ran again after the first launch. Removed the gate — seeding is idempotent (`getAllPresetPhrases()` includes soft-deleted rows, so a user-deleted preset is not resurrected), and this is also what reconciles stored phrases against the arrays. A phrase *added* to an array in a past release never reached existing users either; that's fixed by the same change.
- `ensurePopulated()` now resyncs `sort_order` from the array index for rows that already exist.
- Editing a preset soft-deletes its `PresetPhrase` row and stores a "shadow" `Phrase` under the same id, so shadows carry their own `sort_order` and are resynced independently of the preset row — otherwise an install that had already resynced its preset rows would strand its shadows at the pre-reorder index.
- `getPhrasesForCategoryFlow` now sorts by `sortOrder`. Neither DAO query has an `ORDER BY`; `PresetsViewModel` sorted on its way to the grid but `EditCategoryPhrasesState` did not, so after a resync the Presets grid and the Edit Phrases list disagreed on order.

Known limitation, documented in `arrays.xml`: the resync handles additions and reorders, not removals or moves between arrays, since it never deletes or re-parents an existing row.

### Tests

- `PhraseGridResourceInvariantTest` — asserts `columns × rows == max_phrases` for both grids at every breakpoint, which is the invariant #611 violated.
- `KeypadPhraseOrderTest` — pins the array declaration order and asserts the full rendered row matrix against the signed-off layouts, so a column count with no approved layout fails loudly.
- Both discover breakpoint dirs from disk rather than a hardcoded list, so a new `values-*` dir is covered automatically.
- `RoomPresetPhrasesRepositoryTest` — covers reseeding, the `sort_order` resync, shadow resync, and that soft-deleted presets stay deleted across launches.

Closes #611.

## Test plan

- [x] `./gradlew :app:testDebugUnitTest` — 40 tests, all passing
- [x] `./gradlew :app:compileDebugAndroidTestKotlin` compiles
- [x] Tablet portrait/landscape Compose previews render 3×4 with no blank row
- [ ] `./gradlew connectedDebugAndroidTest` on a device/emulator (not run locally — no AVD available)
- [ ] Tablet AVD (`sw600dp`+), portrait + landscape: "123" renders the 3×4 keypad
- [ ] Handset portrait/landscape: "123" renders 3×4 / 6×2, other categories unchanged
- [ ] **Upgrade path** — install a pre-change build, open it, then install this build: "123" shows keypad order in both the Presets grid and Settings → Edit Categories → 123
- [ ] Edit a "123" phrase (creating a shadow), then relaunch: the edited phrase keeps its keypad position
- [ ] Delete a preset phrase, then relaunch: it stays deleted

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/src/main/java/com/willowtree/vocable/ui/presets/PresetsScreen.kt
+++ b/app/src/main/java/com/willowtree/vocable/ui/presets/PresetsScreen.kt
@@ -679,19 +679,18 @@ fun PresetsScreenPreview2() {
             ),
             selectedCategory = Category.PresetCategory(PresetCategories.USER_KEYPAD.id, 0, false),
             currentPhrases = listOf(
-                PhraseGridItem.Phrase("1", "0"),
-                PhraseGridItem.Phrase("2", "1"),
-                PhraseGridItem.Phrase("3", "2"),
-                PhraseGridItem.Phrase("4", "3"),
-                PhraseGridItem.Phrase("5", "4"),
-                PhraseGridItem.Phrase("6", "5"),
-                PhraseGridItem.Phrase("7", "6"),
-                PhraseGridItem.Phrase("5", "7"),
-                PhraseGridItem.Phrase("6", "8"),
-                PhraseGridItem.Phrase("7", "9"),
-                PhraseGridItem.Phrase("8", "Yes"),
-                PhraseGridItem.Phrase("9", "No"),
-                PhraseGridItem.AddPhrase
+                PhraseGridItem.Phrase("1", "1"),
+                PhraseGridItem.Phrase("2", "2"),
+                PhraseGridItem.Phrase("3", "3"),
+                PhraseGridItem.Phrase("4", "4"),
+                PhraseGridItem.Phrase("5", "5"),
+                PhraseGridItem.Phrase("6", "6"),
+                PhraseGridItem.Phrase("7", "7"),
+                PhraseGridItem.Phrase("8", "8"),
+                PhraseGridItem.Phrase("9", "9"),
+                PhraseGridItem.Phrase("10", "0"),
+                PhraseGridItem.Phrase("11", "No"),
+                PhraseGridItem.Phrase("12", "Yes")
             ),
             isLoadingPhrases = false,
             onCategorySelected = {},
@@ -700,7 +699,7 @@ fun PresetsScreenPreview2() {
             onNavigateToSettings = {},
             onPhraseClick = { _, _ -> },
             isSpeaking = true,
-            activeText = "0"
+            activeText = "1"
         )
     }
 }
--- a/app/src/main/java/com/willowtree/vocable/ui/presets/PresetsScreen.kt
+++ b/app/src/main/java/com/willowtree/vocable/ui/presets/PresetsScreen.kt
@@ -117,7 +117,11 @@ private fun PresetsContent(
             integerResource(id = R.integer.phrases_columns_one_liner_phrases)
         } else integerResource(id = R.integer.phrases_columns)
     }
-    val phraseRows = integerResource(id = R.integer.phrases_rows)
+    val phraseRows = selectedCategory?.categoryId.let { id ->
+        if (id == PresetCategories.USER_KEYPAD.id) {
+            integerResource(id = R.integer.phrases_rows_one_liner_phrases)
+        } else integerResource(id = R.integer.phrases_rows)
+    }
     val maxCategories = integerResource(id = R.integer.max_categories)
     val maxPhrases = selectedCategory?.categoryId.let { id ->
         if (id == PresetCategories.USER_KEYPAD.id) {
@@ -653,6 +657,16 @@ fun PresetsScreenPreview() {
 @Preview(
     name = "Landscape",
     device = "spec:width=800dp,height=400dp,dpi=240",
+    showBackground = true,
+)
+@Preview(
+    name = "Tablet Portrait",
+    device = "spec:width=800dp,height=1280dp,dpi=240",
+    showBackground = true,
+)
+@Preview(
+    name = "Tablet Landscape",
+    device = "spec:width=1280dp,height=800dp,dpi=240",
     showBackground = true,
 )
 @Composable
--- a/app/src/main/res/values-land/integers.xml
+++ b/app/src/main/res/values-land/integers.xml
@@ -7,6 +7,7 @@
     <integer name="phrases_columns">3</integer>
     <integer name="phrases_columns_one_liner_phrases">6</integer>
     <integer name="phrases_rows">2</integer>
+    <integer name="phrases_rows_one_liner_phrases">2</integer>
     <integer name="settings_options_rows">3</integer>
     <integer name="settings_options_columns">2</integer>
     <integer name="edit_keyboard_columns">10</integer>
--- a/app/src/main/res/values-sw600dp-land/integers.xml
+++ b/app/src/main/res/values-sw600dp-land/integers.xml
@@ -5,8 +5,9 @@
     <integer name="max_phrases_one_liner">12</integer>
     <integer name="max_edit_phrases">6</integer>
     <integer name="phrases_columns">3</integer>
-    <integer name="phrases_columns_one_liner_phrases">6</integer>
+    <integer name="phrases_columns_one_liner_phrases">3</integer>
     <integer name="phrases_rows">3</integer>
+    <integer name="phrases_rows_one_liner_phrases">4</integer>
     <integer name="edit_phrases_columns">2</integer>
     <integer name="edit_phrases_rows">3</integer>
     <integer name="selection_modes_rows">2</integer>
--- a/app/src/androidTest/java/com/willowtree/vocable/CategoriesUseCaseTest.kt
+++ b/app/src/androidTest/java/com/willowtree/vocable/CategoriesUseCaseTest.kt
@@ -39,6 +39,7 @@ class CategoriesUseCaseTest {
     private val storedCategoriesRepository = RoomStoredCategoriesRepository(database)
     private val presetPhrasesRepository = RoomPresetPhrasesRepository(
         database.presetPhrasesDao(),
+        database.phraseDao(),
         dateProvider
     )
     private val storedPhrasesRepository = RoomStoredPhrasesRepository(
--- a/app/src/androidTest/java/com/willowtree/vocable/PhrasesUseCaseTest.kt
+++ b/app/src/androidTest/java/com/willowtree/vocable/PhrasesUseCaseTest.kt
@@ -36,6 +36,7 @@ class PhrasesUseCaseTest {
     private val dateProvider = FakeDateProvider()
     private val presetPhrasesRepository = RoomPresetPhrasesRepository(
         presetPhrasesDao = database.presetPhrasesDao(),
+        phraseDao = database.phraseDao(),
         dateProvider = dateProvider,
     )
     private val storedPhrasesRepository = RoomStoredPhrasesRepository(database, dateProvider)
--- a/app/src/androidTest/java/com/willowtree/vocable/presets/RoomPresetPhrasesRepositoryTest.kt
+++ b/app/src/androidTest/java/com/willowtree/vocable/presets/RoomPresetPhrasesRepositoryTest.kt
@@ -6,9 +6,10 @@ import androidx.test.core.app.ApplicationProvider
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.willowtree.vocable.domain.model.PresetCategories
 import com.willowtree.vocable.domain.model.PresetPhrase
+import com.willowtree.vocable.core.locale.LocalesWithText
 import com.willowtree.vocable.data.repository.RoomPresetPhrasesRepository
+import com.willowtree.vocable.data.room.PhraseDto
 import com.willowtree.vocable.data.room.PresetPhraseDto
-import com.willowtree.vocable.data.room.PresetPhrasesDao
 import com.willowtree.vocable.data.room.VocableDatabase
 import com.willowtree.vocable.utility.FakeDateProvider
 import com.willowtree.vocable.utility.VocableKoinTestRule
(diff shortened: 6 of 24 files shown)


app/src/main/java/com/willowtree/vocable/ui/presets/PresetsScreen.kt
```diff
@@ -117,7 +117,11 @@ private fun PresetsContent(
             integerResource(id = R.integer.phrases_columns_one_liner_phrases)
         } else integerResource(id = R.integer.phrases_columns)
     }
-    val phraseRows = integerResource(id = R.integer.phrases_rows)
+    val phraseRows = selectedCategory?.categoryId.let { id ->
+        if (id == PresetCategories.USER_KEYPAD.id) {
+            integerResource(id = R.integer.phrases_rows_one_liner_phrases)
+        } else integerResource(id = R.integer.phrases_rows)
+    }
```
Clstroud, line 124: This is the right fix for the ticket. Worth naming the underlying cause though, because otherwise it comes back.

`phraseRows`, `phraseColumns`, and `maxPhrases` are three independent resources that have to be kept in sync by hand. This PR adds the fourth override that was missing — but nothing prevents the fifth from going missing later.

Compare iOS, where page size is *derived* rather than declared — `CarouselGridLayout.swift`:

```swift
var itemsPerPage: Int { rowCount * columnCount }
```

The invariant `max_phrases == rows × columns` is real in this codebase but unwritten and unenforced. The old tablet-landscape values were 6 cols × 3 rows = 18 cells against a 12-item page — exactly one blank row of 6, which is the bug you fixed. The other direction is worse and silent: if `rows × columns < maxPhrases`, phrases get paged in and never rendered at all, since the grid below renders a fixed `phraseRows × phraseColumns` matrix and indexes into the page.

So as a **follow-up PR**, not this one:

```kotlin
val maxPhrases = phraseRows * phraseColumns
```

and delete `max_phrases` / `max_phrases_one_liner` entirely.

Separately, and small: now that three consecutive `val`s branch on the same condition, hoisting it reads better and makes "have I covered every layout param?" obvious at a glance.

```kotlin
val isKeypad = selectedCategory?.categoryId == PresetCategories.USER_KEYPAD.id

val phraseColumns = integerResource(
    if (isKeypad) R.integer.phrases_columns_one_liner_phrases else R.integer.phrases_columns
)
val phraseRows = integerResource(
    if (isKeypad) R.integer.phrases_rows_one_liner_phrases else R.integer.phrases_rows
)
```

I left that as a plain block rather than a one-click suggestion because it spans lines outside this diff hunk. It also drops the `?.categoryId.let { }` pattern — `?.` followed by a non-safe `.let` reads like a typo even when it's deliberate. You inherited that style from the surrounding code, so this is a cleanup of all three sites rather than anything you introduced.

app/src/main/res/values-sw600dp-land/integers.xml
```diff
@@ -5,8 +5,9 @@
     <integer name="max_phrases_one_liner">12</integer>
     <integer name="max_edit_phrases">6</integer>
     <integer name="phrases_columns">3</integer>
-    <integer name="phrases_columns_one_liner_phrases">6</integer>
+    <integer name="phrases_columns_one_liner_phrases">3</integer>
     <integer name="phrases_rows">3</integer>
+    <integer name="phrases_rows_one_liner_phrases">4</integer>
```
Clstroud, line 10: These lines are the whole invariant in one place: `max_phrases_one_liner` (12, line 5) has to equal `phrases_columns_one_liner_phrases` × `phrases_rows_one_liner_phrases` — 3 × 4. It does now. Before this PR it was 12 vs 6 × 3 = 18, hence the blank row.

The issue's acceptance criteria asked for a unit test covering row-count resolution and that box is still unchecked — nothing under `app/src/test/` references these resources today. Could you add one here? Something that walks every breakpoint × category combination and asserts:

```
maxPhrases == phraseColumns * phraseRows
```

That would have caught this bug, and it'll catch the next person who adds a breakpoint or a one-liner category and forgets one of the four values.

app/src/main/res/values-land/integers.xml
```diff
@@ -7,6 +7,7 @@
     <integer name="phrases_columns">3</integer>
     <integer name="phrases_columns_one_liner_phrases">6</integer>
     <integer name="phrases_rows">2</integer>
+    <integer name="phrases_rows_one_liner_phrases">2</integer>
```
Clstroud, line 10: Consistency nit: `phrases_columns_one_liner_phrases` is declared in six qualifier dirs including `values-sw400dp` and `values-sw400dp-land`, but the new `phrases_rows_one_liner_phrases` is only in four.

It resolves correctly today — I traced it, and `values-land` = 2 is what a ~411dp phone in landscape picks up, which matches `phrases_rows` there — but it works by *fallback* rather than by declaration. Whoever next edits `values-sw400dp-land` and sees a columns override with no matching rows override is going to have to redo that same trace.

Worth declaring it in `values-sw400dp` (4) and `values-sw400dp-land` (2) so the two resources always travel as a pair.

Clstroud (commented):
Nice work on this — I checked it against the iOS implementation and the values you landed on are an exact match across all four breakpoints.

For reference, `NumericCategoryContentViewController.swift` sets:

| iOS size class | cols × rows |
|---|---|
| `hRegular_vRegular` (iPad, both orientations) | 3 × 4 |
| `hCompact_vRegular` (iPhone portrait) | 3 × 4 |
| `*_vCompact` (iPhone landscape) | 6 × 2 |

That lines up 1:1 with `values-sw600dp` / `values-sw600dp-land` / `values` / `values-land` after this change. The non-obvious case is tablet landscape — iPad landscape is still `hRegular_vRegular` on iOS, so 3×4 is correct there and the `6 → 3` column change is right.

I also traced resource resolution for the untouched `values-sw400dp*` dirs: `smallestWidth` outranks `orientation` in Android's qualifier precedence, so a ~411dp phone in landscape resolves the new row resource to `values-land` = 2, matching the existing `phrases_rows`. No handset regression — that claim in your description holds up.

Three comments inline. None of them block the fix itself, and the structural one is genuinely a separate PR. The one I'd like addressed here is the unit test.

The tablet `@Preview` specs are a good addition — right instinct for a Compose codebase, and they directly solve the "you can't see this on a phone emulator" problem from the issue.

Clstroud (approved):
I think we're good now! Nice work 💯

kaylaheath-ux:
Screenshot of ordering after the change:

image

Issue #617, opened 2026-07-28
Day 2 — Capture and persist the device's voice when nothing is selected in Vocable yet


rhyslutsky:
##Resources: 
- **Design**:https://www.figma.com/design/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=6539-3337&t=MyWxM81xsZ0ij6GW-0
- **Branch this work builds on: https://github.com/willowtreeapps/vocable-android/tree/feature/voice-selection**
- **Associated Documentation (check here for updates)**: https://drive.google.com/file/d/1XOgeA1krQmhT9xYDu4PbtXeTr5YFnFCj/view?usp=drive_link
- **Spike [doc](https://drive.google.com/file/d/1p2A8BOqCaMPb_Gqpj6WdshwglTjpFz85/view?usp=drive_link)**

---

**To Do:**

1. **Settings root — add the Voice entry point.**
   - Add a fourth `OptionItem` for "Voice" to `SettingsScreen.kt`'s `options` list (today only Edit Categories, Timing Sensitivity, and Selection Mode) — per [[Settings root v2, node 6561:3335](https://www.figma.com/design/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=6561-3335)](https://www.figma.com/design/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=6561-3335); see also the parent grouping section, [[node 6539:3337](https://www.figma.com/design/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=6539-3337&m=dev)](https://www.figma.com/design/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=6539-3337&m=dev), for all deliverables in context. On click, navigates to the Settings → Voice screen (Step 2 below), not directly to Change Voice.

2. **Build the new "Settings → Voice" screen.**
[...]

Issue #613, opened 2026-07-27
Feature: Voice Selection — Hybrid (Device Default + In-App Override)


rhyslutsky:
## Why

Vocable's TTS pipeline has no defined behavior for which voice it speaks with — it relies on whatever `android.speech.tts.TextToSpeech` happens to default to, never reading the device's actual configured voice. Some users also want Vocable to speak in a voice different from the device-wide default (similar to how Google Maps lets you pick a separate navigation voice). This epic tracks bringing Android to a "hybrid" model: live-check the device's default voice when nothing's explicitly chosen, but let an explicit in-app pick persist and stick across restarts.

## Scope

Epic tracking issue — the actual work is broken into the child issues below, each independently scoped and PR'd against `feature/voice-selection`.

## Child Issues

- [x] #630 — Remove Language Selection/FuelIX scaffolding (unrelated work tangled into the same branch)
- [x] #632 — TTS live device-voice fallback + explicit-pick persistence + stale-voice silent revert
- [x] #642 — Fix stale-voice detection to check download status, not just name/locale match (follow-up to #632)
- [x] #634 — Spike: on-device TTS voice-engine validation (2+ real engines)
- [x] #618 — Hide undownloaded/unavailable voices from the picker
- [x] #622 — Design: Figma deliverables for the three UI screens below
- [x] #635 — Settings root Voice row + Settings → Voice screen
- [x] #636 — Change Voice screen visual rework (title, selected-state checkmark, Default-row removal, pagination, empty state)
- [x] #643 — Change Voice: play/preview button per row
- [x] #644 — Change Voice: responsive 2-column grid matching iOS size classes
- [x] #637 — Remove the old Voice entry point from Selection Mode
- [x] #620 — Final regression + QA pass, once everything above lands

## Closed / Superseded

These originally tracked pieces of this epic's scope, now closed with their work folded into the list above — kept here so the history isn't lost:

- #617 ("Day 2 — capture and persist the device's voice") — closed; its TTS-logic scope shipped via #632, its UI scope split into #635/#636/#637
- #619 ("Day 4 — silently revert stale voice") — closed; shipped via #632/#633 + #642/#645 (a real gap #619 didn't originally anticipate — a voice can be uninstalled without disappearing from `getVoices()`); its multi-engine verification AC folded into #634
- #615/#616/#624 (Day 1 spikes) — closed, consolidated into #634

## Out of Scope

[...]

PR #633, opened 2026-08-03
TTS: live device-voice fallback when nothing is explicitly selected

### LINKED ISSUE #632: TTS: live device-voice fallback when nothing is explicitly selected
Part of #617.

## Why

`VocableTextToSpeech.applySelectedVoice()` (`app/src/main/java/com/willowtree/vocable/core/VocableTextToSpeech.kt:129-141`) early-returns when no voice is explicitly selected — Vocable never reads or applies the device's actual configured TTS voice in that case, it just leaves whatever the engine's internal state happens to be at init. `getDefaultVoice()`/`getVoice()` are never called anywhere in the file today. The existing "Default" row in the picker correctly clears the persisted preference but doesn't make Vocable actively reflect the device's live voice afterward.

## Scope

Make `VocableTextToSpeech`'s voice-resolution logic (used by every `speak()` call) live-check the device's default voice when nothing is explicitly selected, and silently fall back to that same check if a persisted explicit selection stops resolving.

## Acceptance Criteria

- [ ] No explicit selection → `speak()` reads the device's live current voice via `getDefaultVoice()` each time it's needed, never persisted
- [ ] Explicit selection exists and resolves → used as today (no change needed)
- [ ] Explicit selection exists but no longer resolves (voice removed/unavailable) → clear the persisted value and silently fall back to the live check, no user-facing notification
- [ ] Voice application runs *after* `tts.setLanguage()` on every `speak()` call, since `setLanguage()` clobbers a previously-applied voice; the existing `lastSetLocale` skip (line 104) stays in place
- [ ] All three existing call sites behave correctly: `MainActivity.kt:114`, `PresetsViewModel.kt:125`, `KeyboardViewModel.kt:121`
- [ ] Gaze/dwell timing (`GazeClickable`, `isSpeakingFlow`) unaffected
- [ ] Verified on-device/emulator: setting the system TTS voice via `adb shell` and confirming the live fallback path actually reflects it

## Out of Scope

- Settings root "Voice" row, the new Settings → Voice screen, and the Change Voice picker's visual rework — tracked in #617
- Removing the Voice row from `SelectionModeScreen.kt` — tracked in #617
- Play/preview icon behavior — tracked in #617, blocked on sample-phrase confirmation per #617's own notes

## Links

- Parent: #617

MicGon7:
Part of #617 — closes #632.

> [!NOTE]
> **Based on `feature/voice-selection`, not `main`** — same reasoning as PR #631: this is a sub-issue PR for #617/#613, and all of them target `feature/voice-selection` until the whole feature is ready to ship.

## What

- `VocableTextToSpeech.speak()` now returns `Boolean` (`true` = the caller's persisted voice selection was stale and should be cleared). All three call sites (`MainActivity.kt`, `PresetsViewModel.kt`, `KeyboardViewModel.kt`) updated to clear it via `setSelectedVoiceName(null)` when that happens.
- `applySelectedVoice()` now applies the device's live default voice (via `getDefaultVoice()`, checked fresh every call, never persisted) when nothing's explicitly selected, instead of no-op-ing as it did before.
- Extracted the actual decision logic into a pure `resolveVoiceSelection()` function, kept free of `android.speech.tts` types specifically so it's unit-testable — this module's tests use the plain Android SDK stub jar (no Robolectric, no mocking framework), under which real `Voice`/`TextToSpeech` objects can't be constructed. 5 new tests in `VocableTextToSpeechTest.kt` cover all three branches (live default / explicit match / stale fallback).
- `lastSetLocale` skip logic and voice-after-`setLanguage()` ordering preserved unchanged.
- Preserved `feature/voice-selection`'s pre-removal state as branch `archive/language-selection-pre-removal` (commit `b450b519`) — #617's own acceptance criteria required Language Selection be kept on a branch rather than deleted outright, and PR #631 had already merged before this was caught.
- Added a `create-ticket` skill (`.claude/skills/create-ticket/SKILL.md`) and expanded `CLAUDE.md`'s convention into a full "Starting new work" standard (issue-first, `feature/<issue>/<description>` branch naming, `Documentation/work-log/` entries) — directly aimed at preventing tickets from growing into the kind of sprawling documents #613/#617/#622 became.
- `Documentation/work-log/632-tts-live-voice-fallback.md` is the first entry under that new standard.

## Why it's safe

- `./gradlew clean assembleDebug assembleDebugAndroidTest testDebug` succeeds — matches CI's exact task list
- 43 unit tests, 0 failures (5 new, covering the extracted pure decision logic)
- App installs and launches on an emulator without crashing; the pre-existing locale-check path (unrelated to this change) still behaves identically

## Out of scope / follow-up

- **Known verification gap**: the available emulator has no TTS engine installed at all and no Play Store to add one, so the actual `getDefaultVoice()` live-fallback behavior could only be verified via unit tests + a no-crash check, not #617's full manual-test plan (confirming spoken voice matches a manually-set system voice, live-check reflects a mid-session system-voice change, explicit picks stay sticky). **Needs a real device or Play-Store-enabled emulator before this is considered fully verified** — flagging clearly rather than asserting it's done.
- Settings root row, Settings → Voice screen, Change Voice visual rework, removing Voice from `SelectionModeScreen.kt`, and preview/play-icon work all remain tracked in #617.


app/src/main/java/com/willowtree/vocable/ui/presets/PresetsViewModel.kt
```diff
@@ -122,11 +122,12 @@ class PresetsViewModel(
             }
             is PresetsIntent.Speak -> {
                 _state.update { it.copy(activeText = intent.text) }
-                VocableTextToSpeech.speak(
+                val selectionWasStale = VocableTextToSpeech.speak(
                     locale = null,
                     text = intent.text,
                     selectedVoiceName = sharedPreferences.getSelectedVoiceName()
                 )
+                if (selectionWasStale) sharedPreferences.setSelectedVoiceName(null)
```
Mansimran-Singh, line 130: Just curiosity, why are we setting shared prefs to null when it is stale? are we updating it later to non-stale?
or if we have plan to handle it future, let's add a TODO to not forget about it?

app/src/main/java/com/willowtree/vocable/MainActivity.kt
```diff
@@ -110,12 +110,14 @@ class MainActivity : ScopeActivity() {
...
                             when (event) {
-                                is FaceTrackingEvent.Speak ->
-                                    VocableTextToSpeech.speak(
+                                is FaceTrackingEvent.Speak -> {
+                                    val selectionWasStale = VocableTextToSpeech.speak(
                                         locale = Locale.getDefault(),
                                         text = event.text,
                                         selectedVoiceName = sharedPrefs.getSelectedVoiceName()
                                     )
+                                    if (selectionWasStale) sharedPrefs.setSelectedVoiceName(null)
```
Mansimran-Singh, line 119: Duplicate for https://github.com/willowtreeapps/vocable-android/pull/633/changes#r3706608668

app/src/main/java/com/willowtree/vocable/ui/presets/PresetsViewModel.kt
MicGon7: Good question — this is intentional, terminal behavior, not a stopgap, so no TODO needed here.

We never update it back to non-stale automatically. Clearing to `null` *is* the fix: once `selectedVoiceName` is null, every future `speak()` call falls through to the live device-default-voice check added in this same PR (`applySelectedVoice()`'s `VoiceResolution.LIVE_DEFAULT` branch) — so the user immediately gets a working voice again on the very next phrase, with no dead/broken state in between.

The only way a new explicit value gets set again is the user picking a different voice from Vocable's own picker (`VoiceSelectionViewModel.onVoiceSelected()`) — this mirrors the resolved product decision in #613/#619: "clear the persisted value and silently fall back to the live device-voice check ... no notification, toast, or alert shown to the user." So there's nothing pending to implement later; this is the complete, designed behavior for a stale/removed voice.

app/src/main/java/com/willowtree/vocable/MainActivity.kt
MicGon7: Same answer as the other thread: https://github.com/willowtreeapps/vocable-android/pull/633#discussion_r3706962348 — intentional terminal behavior, no TODO needed.

Issue #644, opened 2026-08-03
Change Voice: responsive 2-column grid matching iOS size classes


kaylaheath-ux:
Part of #613.

## Why

iOS's voice picker lays out `fixedCount(2)` columns on tablet and in landscape, dropping to a single column only on phone portrait. Android's picker is single-column everywhere, with `itemsPerPage` hardcoded to 3/5 and padding/button sizes hardcoded in dp — which also violates this repo's convention that grid dimensions come from `integerResource`s per screen-size breakpoint dir.

## Scope

Make the Change Voice picker's grid responsive and column-count-correct per breakpoint, matching iOS's size-class behavior.

## Acceptance Criteria

- [ ] Column count per breakpoint matches iOS's size classes — **1 column** for phone portrait (`values`, `values-sw400dp`, i.e. iOS `hCompact_vRegular`), **2 columns** everywhere else (`values-land`, `values-sw400dp-land`, `values-sw600dp`, `values-sw600dp-land`)
- [ ] Column/row counts exposed as `voice_columns` / `voice_rows` integer resources. Check how the existing `phraseColumns`/`phraseRows` integers are distributed across these dirs and follow the same pattern (defined in every dir vs. only where the value differs from its fallback)
- [ ] Row counts chosen so a full page of rows fits the available height at each breakpoint without clipping or leftover dead space, verified on device. Starting points to validate, not requirements: 5 rows single-column, 3 rows landscape 2-column, 4 rows `sw600dp` 2-column
- [ ] Hardcoded `itemsPerPage`, padding, close-button and paging-button dp values in `VoiceSelectionScreen.kt` replaced with breakpoint resources
- [ ] Layout verified at all six breakpoint dirs, portrait and landscape; fixed tile positions preserved (a deliberate accessibility contract for gaze/motor-impaired users, not a layout preference — do not switch to a fluid or lazy grid)

## Out of Scope

- Title, selected-state checkmark, Default-row removal, pagination, empty state — #636
- Play/preview button — #643
- Note: Figma node `6537:3331` shows tablet landscape as a *single* column, which contradicts iOS. iOS wins per the parent ticket's direction; the divergence is flagged to design on #622

## Links

- Parent: #613
- iOS reference (`updateLayoutForCurrentTraitCollection`): https://github.com/willowtreeapps/vocable-ios/blob/develop/Vocable/Features/Settings/VoiceSettings/VoicePickerViewController.swift

rhyslutsky:
[...]

Issue #643, opened 2026-08-03
Change Voice: play/preview button per row


kaylaheath-ux:
Part of #613.

## Why

iOS's voice picker shows a play/stop button on every voice row that previews that voice before you commit to it. Android has no preview affordance at all. This was previously blocked on confirming a sample phrase; iOS's is now confirmed (`voice_preview.sample_audio.introducion_format` = `"Hello, this is %1$@"`, utterance rate 0.5), which unblocks the work — with one Android-specific wrinkle: Android TTS exposes no friendly voice names, so the name can't be interpolated the way iOS does.

## Scope

Add a per-row play/stop preview button to the Change Voice picker, with working preview playback.

## Acceptance Criteria

- [ ] `VoiceOptionRow` split into two independent gaze targets — a leading play button plus the primary label button — matching iOS's `accessoryButtonStackView` + `primaryLabelButton` structure
- [ ] Play button is square and sized to the row height, matching iOS's `ActionsConfiguration.LayoutSize.square` (`fractionalHeight(1.0)`). Note Figma draws it as 72×72 inside an 88-tall row; iOS's own sizing is the parity target
- [ ] Icon toggles play → stop while a preview is playing; activating it again stops playback
- [ ] Preview speaks via `speak(voice.locale, sample, selectedVoiceName = voice.name)` — plays the voice *without* persisting it as the selection. No `VocableTextToSpeech` changes required
- [ ] `VoiceSelectionViewModel` tracks which voice is previewing and observes `isSpeakingFlow` to clear that state when playback ends
- [ ] Sample phrase added as a non-parameterized `voice_preview_sample` string, wording per product (see comment). It cannot interpolate the voice name: Android's constructed display name yields "Hello, this is English (United States) – Enhanced", whose en-dash (U+2013) is mis-spoken or mis-paused by most engines
- [ ] Play button's dwell visual is not left latched after preview playback ends — `GazeClickable` holds its "selected" state until TTS completes, so verify the button returns to rest on a real device

## Out of Scope

- Final sample-phrase wording — a product copy decision, raised in a comment on this issue; implementation can proceed against a placeholder
- Title, selected-state checkmark, Default-row removal, pagination, empty state — #636
- Responsive 2-column grid — #644
- Any change to `VocableTextToSpeech`'s voice-resolution logic — #632 owns that; preview reuses the existing `speak()` signature as-is

## Links

- Parent: #613
[...]

Issue #637, opened 2026-08-03
Remove Voice entry point from Selection Mode


MicGon7:
Part of #613.

## Why

Voice is moving to Settings root (see #635), but `SelectionModeScreen.kt` still has a live Voice row (`SettingsButton`, `onClick = onVoiceSelection`) and `SelectionModeViewModel` still maintains `selectedVoiceLabel` state wired to it. This needs to be actively removed, not left in place — otherwise there'd be two reachable entry points to the same picker. (Mirrors the Language-row removal already done for this same screen in #630/#631.)

## Scope

Delete the Voice row and its wiring from Selection Mode.

## Acceptance Criteria

- [ ] `SelectionModeScreen.kt`'s Voice `SettingsButton`/row removed, along with the `onVoiceSelection` callback param and its `ConstraintLayout` anchor
- [ ] `SelectionModeViewModel`'s `selectedVoiceLabel` state/refresh logic removed
- [ ] `VocableNavHost.kt`'s wiring of `onVoiceSelection` into `SelectionModeScreen` removed (the `voiceSelection` route itself stays — Settings → Voice still navigates to it, per #635)
- [ ] Selection Mode screen still renders correctly with just Head Tracking after the Voice row is removed (no layout gap)
- [ ] No regression to existing Selection Mode tests

## Out of Scope

- Settings root Voice row / Settings → Voice screen — tracked in #635
- Change Voice visual rework — tracked in #636

## Links

- Parent: #613
- Precedent: #630/#631 (Language row removal from this same screen)

rhyslutsky:
@Mansimran-Singh Please split out this AC in the description above: 
### Product AC: 

- [x] Selection Mode screen still renders correctly with just Head Tracking after the Voice row is removed (no layout gap)

Issue #634, opened 2026-08-03
Spike: on-device TTS voice-engine validation


MicGon7:
Part of #613.

## Why

#632/#633 implemented the live device-voice fallback (`VocableTextToSpeech.applySelectedVoice()` now calls `TextToSpeech.getDefaultVoice()` when nothing's explicitly selected) but flagged an open caveat: `getDefaultVoice()`'s behavior is engine-dependent and may not always reflect the user's actual system-settings choice, depending on the TTS engine installed. This was never validated on a real device — the emulator used during that work had no TTS engine installed at all. Separately, `tts.setLanguage()`'s potential to clobber a previously-applied voice (mitigated via the `lastSetLocale` skip + voice-after-`setLanguage()` ordering) has also only been reasoned about in code, never confirmed on real hardware. #642/#645 later fixed and confirmed the stale-voice-revert path (a voice's data being uninstalled) on one engine (Google's, on a Google Play emulator image) — that repro also needs repeating across engines, folded in here rather than left as a dangling AC on #619.

## Scope

On-device validation of `VocableTextToSpeech`'s voice-resolution behavior across 2+ real TTS engines/devices.

## Acceptance Criteria

- [ ] Confirmed on 2+ engines (e.g. Google Speech Services + one OEM engine like Samsung TTS) whether `getDefaultVoice()` reflects the user's actual system-settings voice choice, or just the engine's own internal default
- [ ] Confirmed `setVoice()`-after-`setLanguage()` ordering actually sticks (doesn't get clobbered) on each tested engine
- [ ] Repeat #642/#645's stale-voice-revert repro (select a downloaded voice, uninstall its data via the OS, confirm silent fallback + preference clears to "Default") on 2+ engines — only verified on Google's engine so far
- [ ] Findings documented, including any per-engine divergence found
- [ ] If a divergence is found that breaks #632/#633's or #642's fallback logic, filed as a new, separately-scoped bug rather than reopening either

## Out of Scope

- Any code changes to `VocableTextToSpeech.kt` — this is a validation-only spike; a fix (if needed) gets its own issue
- Hiding undownloaded voices — tracked in #618

## Links

- Parent: #613
- Prior implementation: #632 / PR #633, #642 / PR #645
- Supersedes the device/engine-matrix AC originally on #619

MicGon7:
**Findings (validated on Google's TTS engine, on-device emulator with a real TTS engine installed — the original blocker at #633's PR time):**

[...]

PR #649, opened 2026-08-04
Change Voice screen visual rework

### LINKED ISSUE #636: Change Voice screen visual rework
Part of #613.

## Why

`VoiceSelectionScreen.kt` works functionally — list, pagination, swipe gestures — but was built against an earlier design: it's titled "Voice" instead of "Change Voice," selection is shown via a plain "Selected" text label, it has a "Default" row with no iOS equivalent, and pagination reuses Presets-screen icons/strings rather than behaving like the shared component every other Vocable paging screen uses. iOS's shipped `VoicePickerViewController` is the source of truth for this screen's design (not the Figma mock, which diverges — see AC below).

## Scope

Bring the Change Voice picker's static visuals, copy, and pagination in line with iOS's shipped voice picker.

## Acceptance Criteria

- [x] Screen title renamed "Voice" → "Change Voice" (iOS `voice_picker.title`); `close_voice_selection` string updated to match
- [ ] Selected state shown as a **trailing checkmark** (iOS `VocableListCellAccessory.checkmark`), replacing the "Selected" text label — not a border. `SelectedColor` (#00FA9A) is already `VocableButton`'s dwell-press color, so a green selected border would read as "being pressed right now"
- [ ] "Default" row removed (`voice_default` string deleted). iOS has no device-default affordance; `setSelectedVoiceName(null)` stays in use by the stale-voice revert in `MainActivity`/`PresetsViewModel`/`KeyboardViewModel`
- [ ] Pagination extracted to a shared `VocablePagination` composable matching iOS's `PaginationView`: horizontally centered (not `SpaceBetween`), and both buttons disabled when there's only one page. Consumed by this screen only in this PR
- [ ] Existing wraparound paging **retained**. iOS's carousel is pseudo-infinite — `CarouselCollectionViewDataSourceProxy` repeats the content 100× and `scrollToMiddleSection` starts the user mid-list, so paging past the last page rolls into the first. Android's `% totalPages` already matches that; an earlier revision of this ticket wrongly called for clamping
- [ ] Pagination buttons get real content descriptions ("Previous page" / "Next page") via `GazeButton`'s `accessibilityLabel`, replacing the `contentDescription = null` the current inline arrows use — gaze hover feeds TTS announcements as a screen-reader substitute on this app, so null labels are a genuine gap
- [ ] `phrases_page_number` reused as-is — its text already matches iOS's `paging_progress_indicator_format` exactly and it's localized in 11 languages
- [ ] Empty state added (iOS `VoicePickerEmptyStateConfiguration`): "No Voices" / "No voices that match your language and region settings are installed." Currently the screen renders blank spacers
- [ ] `VoiceSelectionViewModel` test coverage added (hand-written fake prefs + Turbine) — neither the screen nor its ViewModel has any today
- [ ] Figma divergences flagged back to design on #622: checkmark not 3px border, 2 columns on tablet landscape (node `6537:3331` shows one), and `"Aria (Enhanced)"`-style voice names being unachievable on Android TTS

## Out of Scope

- Play/preview button and tap-to-speak behavior — #643
- Responsive 2-column grid matching iOS's size classes — #644
- Hiding undownloaded voices / removing the download-icon branch in `VoiceOptionRow` — #618. This issue lands *first* and leaves that branch untouched: #618's final AC needs cross-engine confirmation from the #634 hardware spike, so gating visual work on it would stall indefinitely. #618 rebases and deletes the branch it already owns
- Settings root Voice row and Settings → Voice screen — #635. Note iOS's replacement for the download deep-link #618 deletes is a footer on *that* screen: `voice_settings.voice_picker_footer.title`
- Removing the Voice row from Selection Mode — #637. Must land *after* this issue; it's the only entry point available for verifying this screen today
- Voice display-name copy — Android TTS exposes no friendly names, so #622's open "quality-label vocabulary" decision has no iOS counterpart to reconcile against; Android keeps constructed locale + quality

## Links

- Parent: #613
- iOS reference: https://github.com/willowtreeapps/vocable-ios/blob/develop/Vocable/Features/Settings/VoiceSettings/VoicePickerViewController.swift
- Design: https://www.figma.com/design/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=6537-3331

kaylaheath-ux:
Part of #613 — closes #636.

> [!NOTE]
> Targets `feature/voice-selection`, not `main`, per #613's integration-branch workflow. Note that `closes #636` will **not** auto-fire on this merge, since GitHub only auto-closes for PRs targeting the default branch — #636 needs closing by hand afterwards.

## What

- **Title** `voice_selection_title`: "Voice" → **"Change Voice"** (iOS `voice_picker.title`); `close_voice_selection` updated to match.
- **Selected state** is now a trailing **checkmark** (new `ic_check.xml`) rather than the word "Selected", matching iOS's `VocableListCellAccessory.checkmark`.
- **"Default" row removed**, along with the `voice_default` string — iOS has no device-default affordance in its picker.
- **Pagination extracted** to a new shared `ui/components/VocablePagination.kt`: horizontally centered instead of spread `SpaceBetween`, both buttons disabled when there's only one page, and real content descriptions (`previous_page` / `next_page`) replacing `contentDescription = null`.
- **Empty state** added, mirroring iOS's `VoicePickerEmptyStateConfiguration`. Previously an empty voice list rendered as blank spacers.
- **`phrases_page_number` format bug fixed** across all 13 locale files.
- **`VoiceSelectionViewModelTest.kt`** added — the screen and its ViewModel had zero coverage.
- **Work log** at `Documentation/work-log/636-change-voice-visual-rework.md`.

Four commits, separated so the 13-file locale change can be reviewed apart from the screen work.

## Why it's safe

**Build and tests**

```
./gradlew testDebug assembleDebug assembleDebugAndroidTest   → BUILD SUCCESSFUL
52 unit tests, 0 failures, 0 errors   (9 of them new)
app-debug.apk + app-debug-androidTest.apk both produced
```

**The format fix, verified against the real formatter** rather than asserted:

| | Output |
|---|---|
| `String.format("Page %1d of %2d", 1, 6)` | `Page 1 of  6` ← double space |
| `String.format("Page %1$d of %2$d", 1, 6)` | `Page 1 of 6` |
| `String.format("Page %1d of %2d", 1, 10)` | `Page 1 of 10` ← why it hid at ≥10 pages |

Without the `$`, `%2d` means *"decimal, minimum width 2"*, so page counts below 10 were space-padded. Reproduced on device before the fix and confirmed gone in previews after.

**Verified on device** (portrait + landscape, via Settings → Selection Mode → Voice): title reads "Change Voice", no Default row, pagination renders as a centered cluster, paging and swipe both wrap and agree with each other.

**Verified via Compose previews:** `VoiceSelectionScreenPreview` (checkmark on the selected row alongside a download-arrow row), `VoiceSelectionScreenEmptyPreview`, and `VocablePagination{,SinglePage}Preview` (enabled vs. half-opacity disabled arrows).

**Known verification gap — needs a real device.** On the emulator used, every voice reports `isDownloaded == false`, so every row shows a download arrow and tapping routes to `onDownloadVoice()` instead of `onVoiceSelected()`. No voice can be selected there, so the checkmark can't render on device and was confirmed by preview only. Most likely the emulator has no TTS voice data installed (#634 recorded the same during #632); the alternative is that its engine over-reports `KEY_FEATURE_NOT_INSTALLED`, which is exactly what #618/#634 exist to determine. **Someone with real installed voices should confirm the checkmark before #620's QA pass.**

## Design decisions worth a reviewer's attention

**Checkmark, not the 3px border the mock specifies.** iOS uses only a checkmark — `shouldSelectItemAt` returns `false` in `VoicePickerViewController`, so the cell's selection-background path never fires. Separately, a green border would have collided with two existing gaze states: `VocableButton` already uses `SelectedColor` (#00FA9A) as its dwell-press fill and a 4dp `ColorAccent` border on hover.

**Wraparound kept, not clamped.** #636 originally called for clamping; that was corrected on the issue before any code was written. iOS's carousel repeats its content 100× (`CarouselCollectionViewDataSourceProxy.repeatCount`) and starts mid-list, so there's no reachable first or last page. Clamping would also have left the "previous" button enabled-but-inert on page 1 — a full 1000ms dwell that does nothing — since iOS disables purely on `pageCount > 1`, not position.

**Three Figma divergences flagged to design on #622** rather than built: border vs. checkmark, one vs. two columns on tablet landscape, and `"Aria (Enhanced)"`-style voice names, which Android TTS can't produce at all (it returns `en-us-x-tpf-local`, hence the constructed locale + quality label).

## Out of scope / follow-up

- **#643** — play/preview button per row. Unblocked by identifying iOS's sample phrase, but needs a product copy decision since Android can't interpolate a voice name.
- **#644** — responsive 2-column grid matching iOS's size classes. This PR keeps the hardcoded single-column `itemsPerPage = 3/5`, so the screen is **deliberately not yet iOS-complete**.
- **#618** — hides undownloaded voices and deletes the download-icon branch in `VoiceOptionRow`, left untouched here. Ordering was reversed from the original plan: #618's last AC needs cross-engine confirmation from the #634 hardware spike, so gating visual work on it would stall indefinitely.
- **Must land before #637**, which removes the Selection Mode entry point that's currently the only route to this screen.
- **Migrating Presets / EditCategories / EditCategoryPhrases onto `VocablePagination`** — not yet tracked. Kept out to avoid touching the test-pinned `PresetsScreen` in a visual-rework PR.

**Pre-existing issues found but not fixed** (documented in the work log so they aren't misread as regressions): the screen title sits ~36dp right of centre; `MainActivity` pads only horizontal window insets so the bottom system bar draws over content in landscape; and `settings.gradle.kts` still declares the FuelIXLate GitHub Packages repo despite #630 removing that pipeline.


kaylaheath-ux:
Screenshot Evidence

Screenshot 2026-08-04 at 9 36 43 AM
Screenshot 2026-08-04 at 9 38 41 AM
Screenshot 2026-08-04 at 9 47 23 AM
Screenshot 2026-08-04 at 9 47 45 AM

Mansimran-Singh (approved):

MicGon7 (approved):
Looks gravy!

Issue #618, opened 2026-07-28
Hide unavailable/undownloaded voices from the picker


rhyslutsky:
Part of #613.

## Why

`getAvailableVoices()` currently includes undownloaded voices with a download-icon affordance (`onDownloadVoice()` / `ACTION_INSTALL_TTS_DATA` deep-link). Per iOS parity, undownloaded voices should be hidden entirely instead.

## Scope

Filter undownloaded/unavailable voices out of the picker instead of showing a download prompt for them.

## Acceptance Criteria

- [ ] `getAvailableVoices()` filters out voices where `isDownloaded == false`, cross-referencing `isLanguageAvailable(locale)` (`LANG_MISSING_DATA` ⇒ not present) as a fallback, since `KEY_FEATURE_NOT_INSTALLED` is inconsistently populated across OEM engines
- [ ] `isNetworkConnectionRequired` voices treated as unavailable offline
- [ ] Download-icon treatment removed from `VoiceOptionRow`, along with `onDownloadVoice()`/`ACTION_INSTALL_TTS_DATA`
- [ ] Only voices actually available/downloaded on-device appear in the picker; no download icon/prompt/deep-link anywhere on the screen
- [ ] Filter behavior confirmed on 2+ engines/OEMs (see #634) — don't assume Google's TTS engine's behavior generalizes

## Out of Scope

- On-device cross-engine validation itself — tracked in #634
- Change Voice's visual rework (title, selected-state, pagination) — tracked in #636

## Links

- Parent: #613
- Related: #634 (cross-engine validation), #636 (visual rework)

Mansimran-Singh:
Decision is to not remove the un-downloaded ones and logic but just make them not visible for now
cc @kaylaheath-ux

rhyslutsky:
### Product AC

- [ ] Only voices actually available/downloaded on-device appear in the picker.
- [ ] No download icon, prompt, or deep-link exists anywhere in the Voice Selection screen.

@kaylaheath-ux please separate out this AC in the description above. Thanks

Issue #50, opened 2020-03-06
Onboarding


kyliefk:

CarolineLaw:
[Document for this feature](https://docs.google.com/document/d/1ZrXnsRpAIUFDL37UGvEGG4f0CZACfSyt_N25U_7SeD8/edit)

PR #655, opened 2026-08-04
Port throughput-forecast skill to GitHub Projects v2 (Project #50)

### LINKED ISSUE #653: Port throughput-forecast skill to GitHub Projects v2 (Project #50)
## Why
We have a Jira-based `throughput-forecast` skill (Monte-Carlo delivery forecasting from real weekly throughput — no story points) that this repo can't use, since we track work in GitHub Projects (Project #50), not Jira. The skill's own docs describe an explicit "adapt to another tracker" path: only the data-gather step is tracker-specific, and GitHub Projects v2 exposes an equivalent status-change history via its timeline API, so this is a straightforward port rather than a rewrite.

## Scope
Add a new GitHub Projects v2 GraphQL gather script (`gh_fetch.py`) plus an adapted `SKILL.md`/`config.yaml`, reusing the existing `forecast.py`/`forecast-html.py` engine scripts unmodified.

## Acceptance Criteria
- [ ] `gh_fetch.py` produces `done.json`/`open.json` matching `forecast.py`'s existing file-mode envelope
- [ ] Uses a single nested GraphQL query per page (no N+1 per-item timeline calls)
- [ ] `DraftIssue` and `PullRequest` project items excluded by default, with counts logged to stderr
- [ ] `forecast.py`/`forecast-html.py` run unmodified against `gh_fetch.py`'s output
- [ ] `.claude/settings.json` allowlists the three scripts (or explicitly flags the coordination needed with `chore/claude-code-setup`, which has an unmerged settings.json of its own)
- [ ] `Documentation/work-log/` entry written covering the resolutiondate-derivation and DraftIssue/PR-exclusion judgment calls
- [ ] End-to-end run against Project #50 sanity-checked against the live board (item counts reconcile, a spot-checked issue's status history matches)

## Out of Scope
- Porting `run.sh`/`forecast.env.example` (the human quick-start wrapper) — candidate follow-up issue, not yet tracked
- Enabling `--include-prs`/`--include-drafts` by default — needs its own verification that `PullRequest.timelineItems` supports status-changed events the same way `Issue` does

## Links
- https://github.com/orgs/willowtreeapps/projects/50/views/1

rhyslutsky:
## Summary
- Adds `gh_fetch.py`, a GitHub Projects v2 GraphQL gather step for the `throughput-forecast` skill, so it can forecast delivery from Project #50 instead of requiring Jira.
- Reuses `forecast.py`/`forecast-html.py` from the original Jira-based skill **byte-for-byte, unmodified** — both are pure math with no network calls, so only the gather step needed porting.
- Adapted `SKILL.md`/`config.yaml` for GitHub, plus a `.claude/settings.json` allowlist for the three scripts.
- Fixes a bulk-import data-quality trap found during live verification (see work-log) that would otherwise have fabricated a fake mega-throughput week.

## Ticket
Closes #653 — **note**: per this repo's convention, `closes #N` does not auto-fire since this PR targets `feature/voice-selection`, not `main`. Will manually close #653 with a link to this PR once merged.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Refactor
- [ ] Tests
- [ ] CI/CD
- [ ] Documentation

## Testing
- [ ] Unit tests added/updated — N/A, these are one-shot Python data/report-generation scripts, not app code
- [ ] Integration tests added/updated — N/A, same reason
- [x] Manual testing performed — full end-to-end run against the real Project #50 data (see `Documentation/work-log/653-throughput-forecast-github-projects.md` for the verification detail: item-count reconciliation, a hand spot-check of issue #618, and a full `forecast.py`/`forecast-html.py` run producing a real Q1–Q4 report)

## Checklist
- [ ] Tests pass locally (`./gradlew testDebug`) — N/A, no app code (`app/`) touched by this PR
- [x] No API keys or secrets in code — `gh_fetch.py` shells out to `gh api graphql`, reusing whatever `gh auth login` already set up on the machine; no token is read from or written to `settings.json`
- [ ] CLAUDE.md updated (if new pattern introduced) — not updated; this doesn't introduce an app-development pattern

## Two things flagged for reviewer attention

1. **`.claude/settings.json` merge-collision risk**: this branch didn't have one yet; a separate unmerged branch (`chore/claude-code-setup`) has its own with the same shape. Whoever merges second should reconcile the two `permissions.allow` lists rather than let one silently overwrite the other.
2. **Bulk-import artifact in the data itself** (not a bug in this PR, but worth knowing): 141 of this project's 169 "Done" items were bulk-added to the board on 2024-05-28 with no real board history before that — `gh_fetch.py` detects this signature and falls back to `closedAt` for those items instead of treating the import instant as a real completion. Full detail in the work-log.

See `Documentation/work-log/653-throughput-forecast-github-projects.md` for the complete verification writeup.

### DIFF
--- a/.claude/scripts/forecast-html.py
+++ b/.claude/scripts/forecast-html.py
@@ -1,0 +1,503 @@
+#!/usr/bin/env python3
+"""Render a forecast.py JSON dump as a single self-contained HTML report.
+
+Companion to forecast.py:
+  1. python3 forecast.py --done … --open … --json-out forecast.json
+  2. python3 forecast-html.py forecast.json report.html
+
+Takes only file-path args and never touches the network, so it's safe to run
+anywhere and prompt-free when allow-listed by exact path.
+
+Charts are plain HTML/CSS bars (no JS, no external assets) so the file stays a
+single static document safe to open anywhere or attach to chat. Per the dataviz
+method: horizontal magnitude bars with rounded data-ends anchored to a baseline,
+one primary hue for magnitude, a reserved status palette (breach/aging/healthy)
+that ALWAYS ships with an icon + text label + sort order (never color alone), and
+native hover tooltips via SVG/element titles.
+
+To re-skin for your brand, swap the C_* palette constants and the HEADLINE/BODY
+fonts below (and drop in a logo in the header if you want one). The default
+palette is validated for contrast; keep an eye on it if you change the hues.
+
+Usage:  python3 forecast-html.py [in_json] [out_html]
+        (defaults: /tmp/forecast.json → /tmp/forecast_report.html)
+"""
+
+import html
+import json
+import sys
+from pathlib import Path
+
+IN = Path(sys.argv[1] if len(sys.argv) > 1 else "/tmp/forecast.json")
+OUT = Path(sys.argv[2] if len(sys.argv) > 2 else "/tmp/forecast_report.html")
+
+# --- Report palette (swap these hexes for your brand; contrast-validated) ---
+C_RED = "#E10321"       # Core Red
+C_BLUE = "#0B4CDB"      # Core Blue — primary magnitude hue for bars
+C_TEAL = "#5EDEFC"      # Core Teal
+C_TAN = "#FBF6F0"       # Core Tan — page background
+C_DARK = "#041531"      # Core Dark — primary text
+C_TAN_SH1 = "#F2ECE3"   # Tan Shade 1 — note bg, bar tracks
+C_TEAL_T2 = "#DBF8FF"   # Teal Tint 2 — pill bg (AAA 13.75 w/ Blue Shade 3)
+C_BLUE_SH3 = "#06244F"  # Blue Shade 3 — secondary text, pill text
+C_GRAY_T2 = "#D3D8E5"   # Gray Tint 2 — hairlines
+
+# Reserved status palette (validated: contrast >=3:1 on white & tan; the red/amber
+# CVD closeness is in the 8-12 floor band, legal here because every status mark
+# ships an icon + text label + sort order — secondary encoding, never color alone).
+ST_BREACH = "#C21807"    # critical — over the 85% SLE
+ST_WARN = "#B8730E"      # warning — aging (between typical and SLE)
+ST_HEALTHY = "#1F8A4C"   # good — within the typical line
+BAND = {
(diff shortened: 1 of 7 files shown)


troymyree-lang (approved):
Really nice work on this. The gh_fetch.py gather step is clean and safe — argv-list subprocess (no shell), parameterized GraphQL variables, reuses gh auth, no secrets in code — and the bulk-import guard that falls back to closedAt for the 2024-05-28 mass-import is a sharp catch that would otherwise have fabricated a fake throughput spike.

One small accuracy fix before merge: the summary says forecast.py / forecast-html.py are "pure math with no network calls." That's true in file mode (the mode this GitHub port uses), but forecast.py still carries its Jira --live path — load_token() reads an Atlassian Basic-auth token from ~/.claude/settings.json and run_curl() hits Atlassian. It's dormant here (needs an explicit --live --jira-base) and no secret is committed, so it's not a security issue, just dead code in a GitHub-Projects skill. Could you either strip the Jira live path from this copy, or reword the claim to something like "runs in file mode; retains an unused Jira live-mode fallback"? Your call on strip vs. reword.

And +1 on your own two flags — reconciling the settings.json allow list with chore/claude-code-setup on whichever merges second, and double-checking the 7 board statuses against the live board. No other concerns.

rhyslutsky:
Thanks for the review! Went with reword-not-strip (per your call, and matching the "your call" offer) — pushed in 6a0d8699:

- `SKILL.md` and the work-log now say forecast.py runs in **file mode** here (pure math, no network calls in that mode), and explicitly call out that it still carries the source skill's Jira `--live` path (`load_token()`/`run_curl()`) as dead code — never exercised since this port never passes `--live`, kept only so the file stays a straight diff against `dq-documentation` for future upstream fixes.

Your other two flags (settings.json reconciliation with `chore/claude-code-setup`, and double-checking the 7 board statuses) are already called out explicitly in the PR description and work-log — happy to action either further if you want something more concrete than a flag before merge.

troymyree-lang:
Thanks for the quick turnaround — the reword reads accurately now: file-mode / pure-math here, with the Jira --live path clearly labeled as dead code and the rationale for keeping it a byte-for-byte diff against the upstream source. Good call. Merging into feature/voice-selection now.

Reminder for whoever lands chore/claude-code-setup next: reconcile its .claude/settings.json allow list with this one so neither silently overwrites the other.

PR #662, opened 2026-08-04
Change Voice: responsive 2-column grid matching iOS size classes

### LINKED ISSUE #644: Change Voice: responsive 2-column grid matching iOS size classes

kaylaheath-ux:
## Summary

- Column counts now mirror iOS's size classes via `voice_columns` / `voice_rows` integer resources in all six breakpoint dirs — **1 column** for phone portrait, **2 everywhere else**. The runtime `BoxWithConstraints` measurement pass and the hardcoded `itemsPerPage = 3/5` are gone.
- Twelve `voice_*` dimens (margins, spacing, button *and* icon sizes) replace every hardcoded dp in the screen; `VoiceOptionRow` loses its `isLandscape` parameter.
- Grid renders as a fixed row×column nest with `Spacer` placeholders, so a partial last page leaves its trailing slots blank instead of reflowing tile positions.
- Rows take a fixed height matched to the play chip and pack from the top; voice names auto-size to a single un-truncated line.

## Ticket

Part of #613 — closes #644

## Type of Change
- [x] New feature
- [ ] Bug fix
- [x] Refactor
- [x] Tests
- [ ] CI/CD
- [x] Documentation

## Testing
- [ ] Integration tests added/updated

`VoiceGridResourcesTest` (4 tests) pins the breakpoint matrix by parsing `integers.xml` off disk — there's no Robolectric here, so a JVM test can't go through a real `Resources`. It was **mutation-checked**: flipping `values-sw600dp`'s `voice_columns` to 1 fails two assertions as intended.

Verified on device at all six breakpoint dirs by overriding `wm size` / `wm density` / rotation and measuring the real hierarchy via `uiautomator dump`, plus screenshots:

| dir | cols × rows | row height |
|---|---|---|
| `values` / `values-sw400dp` | 1 × 5 | 60dp |
| `values-land` / `values-sw400dp-land` | 2 × 3 | 48dp |
| `values-sw600dp` | 2 × 7 | 80dp |
| `values-sw600dp-land` | 2 × 4 | 64dp |

Also checked: partial last page holds tile origins byte-identical to page 1 with the trailing slot blank; rotating while on page 2 lands on a valid page (`itemsPerPage`-keyed reset plus #618's clamp); names render full and un-truncated at every breakpoint.

## Checklist
- [x] Tests pass locally (`./gradlew testDebug`) — 82 tests, 0 failures; `assembleDebug` + `assembleDebugAndroidTest` clean
- [x] CLAUDE.md updated (if new pattern introduced)

## Reviewer notes — two deliberate divergences

**1. Rows are a fixed chip-matched height, so there IS leftover space at the bottom.** This diverges from the ticket's "no leftover dead space" criterion and from iOS's `.flexible(minHeight:)` stretching. Built to spec first, tiles came out up to 244dp tall on tablet portrait — near-empty boxes around one line of text — and the compact row was preferred on review. Because row height no longer depends on row count, `voice_rows` is now purely slots-per-page.

**2. `values-sw600dp` uses 7 rows, not the 4 the ticket suggested** (the ticket called its numbers "starting points to validate, not requirements"). Row count now only affects how many voices fit per page: 14 on tablet portrait.

Row counts do **not** match iOS's derived 9/3/8/7. iOS computes them as `floor(height / minHeight)` and can afford 64pt rows because its labels are short proper names (`AVSpeechSynthesisVoice.name`); Android TTS has no equivalent, so ours carry a 31-character string. Full reasoning in the work log.

## Known limitation

At large system font scales (≳1.25×) the voice name truncates and loses its trailing `Voice N` index, which is the only thing distinguishing rows. Raised by @-review on Slack ("ensure it is readable regardless of text length and font size") and **not fully solved**: with chip-height rows, one line doesn't fit widthwise (~340dp needed vs ~213dp) and two don't fit heightwise (2 × 48dp in an 80dp row). The three remedies each cost something (grow rows with font scale and shrink the row count; drop to one column, which doesn't help phone portrait; or shorten the name copy — the only one that works everywhere, and already under discussion). Documented rather than silently picked.

Worth knowing for future work here: `uiautomator dump` **cannot** detect truncation — its `text` attribute reports the source string, so an ellipsized label still dumps in full. An automated check read `ellipsized=0/9` while the screen was visibly truncated. Confirm from screenshots.

## Also in this PR

Corrects a stale `CLAUDE.md` bullet claiming the keypad grid shape is pinned by `presets/{KeypadPhraseOrderTest,ResourceXml}.kt`. Neither file exists in `src/test` or `src/androidTest` — that claim would have made it reasonable to skip adding a test on exactly this kind of ticket.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kaylaheath-ux:
image
image

MicGon7 (approved):
👍

PR #663, opened 2026-08-04
Update CODEOWNERS

PaulKlauser:
Being honest with myself, I'm not keeping up with development here, and should no longer be a codeowner.


MicGon7 (approved):
👍

PR #668, opened 2026-08-05
Change Voice: fill the page height instead of leaving a dead band

### LINKED ISSUE #667: Change Voice: fill the page height instead of leaving a dead band below a full page
Part of #613

## Why

#644 (PR #662) made the Change Voice grid responsive per breakpoint but shipped a visible
regression: a **full** page of voices covers only part of the grid area, leaving a band of dead
space above the pager — 287dp on phone portrait, 368dp on tablet portrait. The cause is that #644
switched rows from stretching to a fixed chip-matched height without re-deriving the row counts,
which had been chosen for the stretching model. Retuning those counts would not fix it either:
`sw###dp` qualifiers constrain width and never height, so one count per width bucket has to serve
both an 851dp-tall phone and a 640dp one — dead space on the first, clipped rows on the second.

## Scope

Make a full page of the Change Voice grid consume the available page height at every breakpoint and
device height, without clipping rows.

## Acceptance Criteria

- [ ] A full page of rows fills the grid area at every breakpoint — no dead band above the pager
- [ ] No row is clipped or pushed off the page, including on screens shorter than those the layout
      was tuned on
- [ ] A rendered row is never shorter than the chip-matched height design signed off on
      (60 / 48 / 80 / 64dp per breakpoint)
- [ ] Tile size and column are identical on every page; a short page shares its leftover height with
      the gap under the header rather than putting all of it above the pager, bounded so a sparsely
      filled page does not float mid-screen
- [ ] Grid shape covered by a JVM test, including a guard against reintroducing a fixed row count
- [ ] Verified on device at all four dimens breakpoints, portrait and landscape, and at font scale 2.0

## Out of Scope

- Voice names truncating at large font scales — #644's documented copy-length limitation, tracked on #613
- Landscape pager sitting under the system nav bar (global inset issue) — noted on #636
- Screen title sitting right of centre — noted on #636

## Links

- iOS reference (`CarouselGridLayout.computeRowCount()` derives row counts from a `minHeight` the same way; note its `alignment` defaults to `.top`, which this ticket deliberately diverges from): https://github.com/willowtreeapps/vocable-ios/blob/develop/Vocable/Features/Settings/VoiceSettings/VoicePickerViewController.swift

kaylaheath-ux:
## Summary

- Derive the picker's row count from the measured page height against a new `voice_row_min_height` dimen instead of reading a fixed `voice_rows` integer, and fill rows by weight — a full page now consumes the grid area exactly at every breakpoint. #644 left 287dp of dead space below a full phone-portrait page and 368dp on tablet portrait.
- Retire `voice_rows` from all six `integers.xml` dirs (`voice_columns` unchanged) and rename `voice_row_height` → `voice_row_min_height`, same 60/48/80/64dp values, now a floor rather than an exact height. A rendered row is never shorter than that and never more than one row-pitch taller.
- A short page renders only its occupied rows and insets them from the top by half the leftover — centring the group — capped at half a row pitch, so the gap is shared with the space under the header without a sparse page floating mid-screen.
- Page capacity roughly doubles as a result (8 voices per phone-portrait page instead of 5), so a real device's 30 voices span 4 pages rather than 6.

## Ticket

Part of #613 — closes #667

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Refactor
- [ ] Tests
- [ ] CI/CD
- [ ] Documentation

## Testing
- [ ] Integration tests added/updated

`./gradlew testDebug` — 87 green. Reworked `VoiceGridResourcesTest` (columns + row-min-height matrices, plus a guard that fails if a `voice_rows` integer comes back) and added `VoiceRowCountTest` (reference devices, plus sweeps asserting no row is ever below the design height and no page has room for another row). Both groups were mutation-checked: dropping the `+ rowSpacing` term from `voiceRowCount` and re-adding `voice_rows` each fail as intended.

Verified on an emulator with 9 installed voices at all four `dimens.xml` buckets by measuring the real hierarchy with `uiautomator dump` — in each, the grid's bottom edge plus one section spacing lands on the pager's top edge:

| config (app window) | grid | row height | grid bottom + spacing | pager top |
|---|---|---|---|---|
| phone portrait 393×851dp | 1 × 8 | 69dp | 2101px | 2098px |
| phone landscape 532×393dp | 2 × 4 | 58dp | 924px | 926px |
| tablet portrait 800×1280dp | 2 × 10 | 88dp | 2352px | 2352px |
| tablet landscape 1280×800dp | 2 × 7 | 67.5dp | 1393px | 1392px |

Also checked at `font_scale 2.0` (count unchanged, nothing clipped) and on a short page (phone portrait page 2 inset 40dp = half a pitch; tablet portrait 52dp, against 276dp when the same build centred uncapped).

## Checklist
- [x] Tests pass locally (`./gradlew testDebug`)
- [x] CLAUDE.md updated (if new pattern introduced)

### Two things worth a reviewer's attention

1. **The measurement pass #662's review removed is not what came back.** That was a `BoxWithConstraints` + `LaunchedEffect` pair writing a mutable `itemsPerPage` — a two-pass layout that recomposed with a different capacity a frame after first draw. This is a `BoxWithConstraints` feeding a pure function during composition: no state, no `LaunchedEffect`, no first-frame flash. `voice_columns` is still a resource.
2. **The short-page inset diverges from iOS deliberately, on request.** `CarouselGridLayout` has a `.center` alignment but defaults to `.top`, and `VoicePickerViewController` never overrides it (only `ListeningResponseContentViewController` does), so iOS top-anchors a short voice page. This is also the one place where a tile's vertical position depends on how many voices are on the page, which the fixed-position gaze contract otherwise rules out — the cap is what bounds that shift to half a row rather than half a page. Reasoning is recorded in `Documentation/work-log/667-voice-grid-fill-page.md`.


Mansimran-Singh (approved):
Looks good

PR #670, opened 2026-08-05
Reset: category/phrase data wipe + reseed

Mansimran-Singh:
## Summary
- Added bulk `deleteAll*()` queries to `CategoryDao`, `PhraseDao`, `PresetCategoryDao`, `PresetPhrasesDao` — none of the four had a "delete everything" query before
- Added a public `populateDatabase()` on `RoomPresetCategoriesRepository`, mirroring the one that already existed on `RoomPresetPhrasesRepository`, so reseeding categories can be triggered deterministically instead of only implicitly via flow subscription
- Added `resetToDefaults()` to `ICategoriesUseCase`/`IPhrasesUseCase`. `PhrasesUseCase.resetToDefaults()` wipes stored+preset phrases and re-seeds; `CategoriesUseCase.resetToDefaults()` delegates to `phrasesUseCase.resetToDefaults()` (reusing the same collaborator `deleteCategory` already uses to cascade phrase deletion), then wipes+reseeds categories — so `categoriesUseCase.resetToDefaults()` is the single entry point #640 needs for the reset button
- 7 new androidTest cases: reset after category edit / deletion / custom addition, reset after phrase edit / deletion / custom addition, and Recents clearing after reset

## Ticket
Part of #627 — closes #639

This PR targets `feature/638/reset-preference-wipe` (not `feature/voice-selection` or `main`) to avoid rebase/merge churn while both #638 and #639 are in flight under the same epic. `closes #639` won't auto-fire on merge into a non-default branch — will close manually with a link to this PR once merged.

Full details, including why Recents-clearing needed its own explicit test and a known gap (no Compose UI test yet, since there's no reset button until #640) in `Documentation/work-log/639-reset-category-phrase-reseed.md`.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Refactor
- [x] Tests
- [ ] CI/CD
- [x] Documentation

## Testing
- [x] Unit tests added/updated
- [x] Integration tests added/updated
- [x] Manual testing performed

## Checklist
- [x] Tests pass locally (`./gradlew testDebug`) — also ran the full `connectedDebugAndroidTest` suite on-device (55 tests, 0 failed)
- [x] No API keys or secrets in code
- [ ] CLAUDE.md updated (if new pattern introduced)

PR #669, opened 2026-08-05
Reset: full preference wipe + defaults inventory

Mansimran-Singh:
## Summary
- Inventoried every `IVocableSharedPreferences`-backed setting (`mySayings`, `dwellTime`, `sensitivity`, `headTrackingEnabled`, `selectedVoiceName`, `firstTime`) and confirmed each already has a correct, defined default
- Promoted the existing `clearAll()` full-wipe method onto `IVocableSharedPreferences` so it's reachable through DI like every other preference method, instead of only on the concrete class
- Named the two remaining inline defaults (`DEFAULT_HEAD_TRACKING_ENABLED`, `DEFAULT_FIRST_TIME`) alongside the existing `DEFAULT_SENSITIVITY`/`DEFAULT_DWELL_TIME` constants
- Fixed an inverted `setFirstTime()` in `FakeVocableSharedPreferences` (was setting `true` instead of `false`), caught while writing the reset test
- Added `VocableSharedPreferencesResetTest` — one test per preference, verifying `clearAll()` restores its documented default

## Ticket
Part of #627 — closes #638

Note: since this PR targets `feature/voice-selection` (the Voice Selection integration branch — see below) rather than `main`, GitHub's `closes` won't auto-fire on merge. Will close #638 manually with a link to this PR once merged.

**Why this branch is based on `feature/voice-selection`, not `main`:** `selectedVoiceName` and `firstTime` (and their `IVocableSharedPreferences` methods) only exist on `feature/voice-selection` — voice selection (#613) hasn't merged to `main` yet. #638's acceptance criteria explicitly requires covering voice reset, so this needed to build on top of that branch.

**Resolved voice-fallback decision:** #627's original open PM question (mirror iOS's incidental nil-fallback vs. deliberately restore a default voice) was resolved before this ticket started — per Chris Stroud (Slack, 2026-08-04), an unset `selectedVoiceName` (falling back to system voice) is the correct intended end state. `getSelectedVoiceName()`'s existing `null` default already satisfies this. `Documentation/reset-app-settings-ios-reference.md` still frames this as unresolved and is stale relative to #627's issue body.

Full details in `Documentation/work-log/638-reset-preference-wipe.md`.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Refactor
- [x] Tests
- [ ] CI/CD
- [x] Documentation

## Testing
- [x] Unit tests added/updated
- [ ] Integration tests added/updated
- [ ] Manual testing performed

## Checklist
[...]

Issue #639, opened 2026-08-03
Reset: category/phrase data wipe + reseed


MicGon7:
Part of #627.

## Why

#360 (Convert Preset to Custom Categories) made preset categories user-editable/deletable, with its own acceptance criteria assuming "if any of these categories are edited and the app is reset, the categories shall return to their default state." No such reset logic exists today. iOS's version deletes all Core Data `Phrase`/`Category` rows and re-seeds from bundled presets — Android needs the Room equivalent.

## Scope

Delete and re-seed all Room `Phrase`/`Category` data on reset, exactly reversing any #360-driven edits.

## Acceptance Criteria

- [x] All Room `Phrase` and `Category` rows deleted and re-seeded from the bundled preset source
- [x] Reset exactly reverses any #360-driven edits, deletions, or additions — including user-added categories/phrases, not just edits to presets
- [ ] No orphaned custom categories or phrases survive a reset
- [x] Reset does not require an app reinstall or restart to take effect
- [x] Unit tests cover: reset after a category edit, reset after a category deletion, reset after a custom category/phrase addition
- [x] Instrumented UI test confirms post-reset state matches a fresh install's default category/phrase set exactly

## Out of Scope

- Preference wipe — tracked in a sibling issue under #627
- Settings UI entry point/confirmation dialog — tracked in a sibling issue under #627

## Links

- Parent: #627
- Related: #360

rhyslutsky:
[Michael Gonzalez (@MicGon7)](https://github.com/MicGon7) would you please add the following/ split out these in the description above for demo:

###Product AC:


- [x] Built-in/Preset Categories on Home Screen reset to the 8 default categories (General, Basic Needs, Personal Care, Conversation, Environment, 123, Recents).
- [x] "Recents" tile defaults to empty (Home --> Recents --> "No recently used phrases"
- [x] Built-in/Preset Phrases match original defaults. (Settings --> Categories & Phrases --> ))
- [x] Edited Categories and Phrases - No hidden, deleted, reordered state remains

Issue #638, opened 2026-08-03
Reset: full preference wipe + defaults inventory


MicGon7:
Part of #627.

## Why

iOS's reset is a blanket wipe of every preference key, not a curated subset (see `Documentation/reset-app-settings-ios-reference.md`). Android needs the same model, but first needs a confirmed inventory of every `VocableSharedPreferences`-backed setting and its defined default — a gap Android may have that iOS doesn't, since Swift-level `@AppConfig` defaults exist for every property there.

## Scope

Full, unconditional wipe of all app preferences on reset, with every one confirmed to fall back to a defined default.

## Acceptance Criteria

- [x] Full inventory of every `VocableSharedPreferences`/`IVocableSharedPreferences`-backed setting, confirming each has a defined default to fall back to
- [x] All preference values cleared unconditionally on reset — full wipe, not a curated subset
- [x] Every preference falls back to its defined default on next read after reset (Timing Sensitivity, Selection Mode, Voice — per whichever fallback option PM decides, see #627 — and any others in the current settings surface)
- [x] Unit tests cover reset behavior for each preference category

## Out of Scope

- Category/phrase data reset — tracked in a sibling issue under #627
- Settings UI entry point/confirmation dialog — tracked in a sibling issue under #627

## Links

- Parent: #627
- Reference: `Documentation/reset-app-settings-ios-reference.md`

rhyslutsky:
@MicGon7 would you please add the following/ split out these in the description above for demo: 
### Product AC: 

- [x] 'Head tracking enabled 'is reset to default =boolean = true (Settings --> Selection Mode --> Head tracking is ON)
- [x] Gaze/ cursor sensitivity is reset to default =medium (0.1F) (Settings --> Timing & Sensitivity)
- [x] 'Dwell Time' is reset to default =1000L (Settings --> Timing & Sensitivity --> 'Hover Time' = "1 second")
- [x] 'Selected Voice' is reset to default = device voice (Settings --> Voice --> top row matches voice selected on the device)

PR #671, opened 2026-08-05
Reset: Settings UI entry point + confirmation + accessibility

### LINKED ISSUE #640: Reset: Settings UI entry point + confirmation + accessibility
Part of #627.

## Why

Reset needs a discoverable, safe-to-trigger entry point in Settings — safe meaning a clear confirmation step, since the action is total and irreversible (all custom categories, phrases, and settings erased).

## Scope

Add the "Reset App Settings" row, confirmation dialog, and wire it to the reset logic from the other #627 child issues.

## Acceptance Criteria

- [x] A "Reset App Settings" option is available in Settings, styled consistently with the other existing options, and fully operable via head-tracking/gaze selection and switch input, not just touch
- [x] Selecting it shows a confirmation dialog before any data changes; the dialog plainly states the reset is permanent and total
- [x] The dialog presents both a Cancel and a Confirm action, each fully operable via head-tracking/gaze selection and switch input, not just touch
- [x] Selecting Confirm triggers the reset; selecting Cancel makes no changes
- [ ] Reset behaves correctly if triggered mid-speech/TTS playback, during onboarding, or immediately after first install — no crash, no stuck state
- [ ] No regression in the three existing Settings screen options (Categories and Phrases, Timing and Sensitivity, Selection Mode)

## Out of Scope

- Preference wipe logic — tracked in #638
- Category/phrase data reset logic — tracked in #639

## Links

- Parent: #627
- Depends on: #638, #639

Mansimran-Singh:
## Summary
- Added a "Reset App Settings" row (5th `SettingsButton`) to the Settings screen, reusing the screen's one existing confirmation dialog (extended `ExitDialogType` with `RESET_APP_SETTINGS`) rather than building a new dialog primitive
- Dialog copy matches iOS's shipped wording exactly (pulled from `vocable-ios`'s `Localizable.xcstrings`): body *"Are you sure you want to reset Vocable to default settings? This action cannot be undone."*, Cancel / Reset buttons; reuses the row's own "Reset App Settings" label as the dialog title since iOS's alert has no separate title
- `SettingsViewModel.confirmDialog()`'s new branch calls `categoriesUseCase.resetToDefaults()` (also wipes/reseeds phrases per #639) then `prefs.clearAll()` (#638) — a single call handles the full reset
- **Fixed two real bugs found via manual on-device testing**, not caught by any existing test:
  - Dialog Cancel/Reset text rendered in a washed-out pale mint instead of dark navy — the shared `Typography.labelLarge` hardcodes a text color that wins over the button's intended `LocalContentColor` whenever a `Text()` doesn't set its own `color=`. Fixed at the two affected `Text()` calls (also fixes the same latent bug on the pre-existing Privacy Policy/Contact Developers dialog buttons)
  - Selection Mode's Head Tracking toggle didn't reflect a reset in real time. Root cause was two-layered: `FaceTrackingPermissions` cached its `permissionState` in memory, only updated via its own enable/disable methods (fixed by adding a prefs listener, mirroring `FaceTrackingViewModel`'s existing pattern) — but `VocableSharedPreferences.clearAll()`'s bare `clear()` **never fires `OnSharedPreferenceChangeListener` at all** (Android only notifies for keys explicitly put/removed in an edit, not a bare `clear()`). Fixed by following `clear()` with an explicit rewrite of every listener-observed key back to its default. Added `VocableSharedPreferencesTest` (androidTest) with a regression case for this, since it can't be exercised through the JVM fake
- 6 new `SettingsViewModelTest` cases; `FakeCategoriesUseCase.resetToDefaults()` filled in (was `TODO`)

## Ticket
Part of #627 — closes #640

This PR targets `feature/voice-selection` directly (rather than the individual #638/#639 branches) since both of those already merged into it (#669, #670). `closes #640` won't auto-fire since this isn't `main` — will close manually with a link once merged.

Full details, including the exact investigation behind both bugs above, in `Documentation/work-log/640-reset-settings-ui.md`.

## Type of Change
- [x] New feature
- [x] Bug fix
- [ ] Refactor
- [x] Tests
- [ ] CI/CD
- [x] Documentation

## Testing
- [x] Integration tests added/updated

## Checklist
- [x] Tests pass locally (`./gradlew testDebug`) — also ran the full `connectedDebugAndroidTest` suite on a physical device (53 tests, 0 failed)
- [ ] CLAUDE.md updated (if new pattern introduced)


Mansimran-Singh:
[Screen_recording_20260805_161225.webm](https://github.com/user-attachments/assets/82213b31-a799-44e3-8efb-b54e645f9a32)

MicGon7 (approved):
Nice work 👍

